Issue #4, opened 2018-03-11
Install cmake-cargo-gen from cargo if it's not already installed


AndrewGaspar:

AndrewGaspar:
The generator executable will _not_ be published to crates.io - it will ship with the release of Corrosion.

PR #20, opened 2020-05-18
Automatic generator compile and usage example

yazdan:
- updated readme for example usage project
- compile needed generators on demand

This is related to issue #19 and maybe #4


AndrewGaspar:
Hi @yazdan - thank you for submitting this PR!

I've fleshed out the README quite a bit, and I'm explicitly supporting `add_subdirectory` as a valid method for including the project. Could you take a look and let me know if I've already addressed the concerns this PR is trying to address??

yazdan:
Hi @AndrewGaspar It is better than what I have done. Thanks for update!

PR #21, opened 2020-06-01
Fixed dyn trait object deprecation warnings

westernmagic:


AndrewGaspar:
Huh, I don’t think I ever saw notifications for any of these. Thanks for this!

Issue #57, opened 2020-10-06
Selectively import crates


AndrewGaspar:
There should be a way to filter the imported crates from a workspace.

Something like:
```cmake
corrosion_import_crate(
    MANIFEST_PATH Cargo.toml
    CRATES foo)
```

cc @ogoffart

**EDIT:** Additionally, corrosion_import_crate should only import the crates in the target manifest, not for the whole workspace.

AndrewGaspar:
@ogoffart, RE: https://github.com/AndrewGaspar/corrosion/issues/56#issuecomment-704287757

I don't think there's anything I can do about importing a sub-part of a workspace. In fact, what I might do is just fail out if you try to import a member of a workspace and require that you import the root crate.

ogoffart:
Thanks for filling the issue.

Intuitively, i would have thought that importing a sub-part of a workspace would only import the lib from that specific Cargo.toml.
If that's not possible, having a `CRATES foo bar` to filter the target that i'm interested in would work just fine as well.

AndrewGaspar:
Yeah, that's just the way Cargo works - it will look up in the directory structure to find the "root" crate, and then use that manifest instead. I find it kind of surprising, too, but I can see why they'd want that behavior.

AndrewGaspar:
Well, maybe I should be more precise on that - it will only build the crates in the sub-manifest, even though it will operate in the _context_ of the root crate.

Maybe with that framing it should just import the crates from the sub-crate...

AndrewGaspar:
OK, perhaps I'll only import crates from the target crate, even if it's not the root crate. :P

ogoffart:
Hi Andrew, 
Are you currently working on this?
If not, i would like to help and try to contribute, as this is currently an important feature for the project i'm working on.

Also, when selecting specific crates, i would want to be able to choose features, but the `CRATES` syntax would not allow that.

Imagine I want to build two crates and i want to specify some features.
I can think of a few options:

1. Have a `CRATE` and feature option

```cmake
corrosion_import_crate(
    MANIFEST_PATH ${CMAKE_CURRENT_SOURCE_DIR}/Cargo.toml 
    CRATE foo 
    FEATURES ffi
)
corrosion_import_crate(
    MANIFEST_PATH ${CMAKE_CURRENT_SOURCE_DIR}/Cargo.toml 
    CRATE bar 
    FEATURES ${BAR_FEATURES})
)
```

[...]

PR #67, opened 2020-10-09
Add a way to select crates

ogoffart:
Add a `CRATES` option to `corrosion_import_crate` to specify which crates to import

Issue #57


AndrewGaspar:
Looks good - I might do a separate PR to do the path scoped inclusion. Are you going to implement the `CORROSION_FEATURES` property separately?

PR #68, opened 2020-10-09
Use JOB_POOL to avoid running cargo in parallel

ogoffart:
Ninja supports the concept of job pools so that cargo is not run in parallel.
Otherwise, we see a lot of
```
Blocking waiting for file lock on build directory
```


AndrewGaspar:
Oh, good call. I didn't realize you could create granular job pools like that. Re-running the tests because a couple failed due to infrastructure issues (GitHub Actions kind of drives me crazy sometimes), but otherwise looks good.

Issue #65, opened 2020-10-08
Output crate artifacts to standard locations


AndrewGaspar:
RE: https://github.com/AndrewGaspar/corrosion/issues/60#issuecomment-705529748

Currently bins and libs are just left in the cargo build tree, which is out of sync with where CMake-native artifacts are placed. Artifacts should be placed in CMAKE_CURRENT_BINARY_DIR. Placement should also respect "OUTPUT_DIRECTORY" properties:
- [ARCHIVE_OUTPUT_DIRECTORY](https://cmake.org/cmake/help/latest/prop_tgt/ARCHIVE_OUTPUT_DIRECTORY.html)
- [LIBRARY_OUTPUT_DIRECTORY](https://cmake.org/cmake/help/latest/prop_tgt/LIBRARY_OUTPUT_DIRECTORY.html)
- [RUNTIME_OUTPUT_DIRECTORY](https://cmake.org/cmake/help/latest/prop_tgt/RUNTIME_OUTPUT_DIRECTORY.html)
- [PDB_OUTPUT_DIRECTORY](https://cmake.org/cmake/help/latest/prop_tgt/PDB_OUTPUT_DIRECTORY.html)
- [COMPILE_PDB_OUTPUT_DIRECTORY](https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_PDB_OUTPUT_DIRECTORY.html)

kjeremy:
I could try to tackle this with some guidance if it's relatively straightforward.

AndrewGaspar:
Roughly, you'll need to do some updates to [`_add_cargo_build`](https://github.com/AndrewGaspar/corrosion/blob/bf3af38e4f404525af92c381fe81394617b67087/cmake/Corrosion.cmake#L64)

- You'll need to add a `${CMAKE_COMMAND} -E copy` call in the `add_custom_target` call towards the end of the function to move the build artifacts to the target location from the cargo build tree - by default I _think_ this would be ${CMAKE_CURRENT_BINARY_DIR} for all target types
- You'll need to use the overloaded target output directories (listed above) if they're set in place of CMAKE_CURRENT_BINARY_DIR
[...]

PR #69, opened 2020-10-11
Copy cargo artifacts to CMAKE_CURRENT_BINARY_DIR

kjeremy:
Towards #65


kjeremy:
It looks like I'm confusing ninja here. I thought that maybe adding `DEPENDENCIES ${build_byproducts}` would help but it did not.

AndrewGaspar:
Yeah, the current problem that you're running into with Visual Studio builds is because you're not copying the files for multi-config generators. While `BYPRODUCTS` is not possible to express for a multi-config generator, you can at least still do config-selected output destinations.

I can look into it.

kjeremy:
Thank you. That was more complicated than I expected!

PR #73, opened 2020-11-17
Change install path from /usr/Corrosion to /usr

2xsaiko:
cmake should be able to find the cmake module if installed normally using make install.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,7 @@ if (NOT CORROSION_IS_SUBDIRECTORY)
 
     corrosion_install(
         TARGETS corrosion-generator
-        DESTINATION Corrosion/libexec
+        DESTINATION libexec
     )
 
     # Generate the Config file
@@ -62,7 +62,7 @@ if (NOT CORROSION_IS_SUBDIRECTORY)
 
     configure_package_config_file(
         cmake/CorrosionConfig.cmake.in CorrosionConfig.cmake
-        INSTALL_DESTINATION Corrosion/lib/cmake/Corrosion
+        INSTALL_DESTINATION lib/cmake/Corrosion
     )
 
     if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.13")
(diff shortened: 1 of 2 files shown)


AndrewGaspar:
This has been brought up before... https://github.com/AndrewGaspar/corrosion/issues/52

The install path that is used by Corrosion should be found by CMake when calling `find_package(Corrosion)`, without changing any default install paths or environment variables or anything. Could you explain in more detail what the exact problem you're running into is and maybe I can help you debug?

2xsaiko:
Hmm, maybe I've done something wrong when installing since I didn't do it with just make install, instead writing an ebuild for it.

Still though, /usr/Corrosion is not a standard install path and it most likely won't be officially packaged like that in any distro. (I just noticed because portage yelled at me that the package installed into non-standard paths.) Why is it installing to there, anyways? Any other cmake plugin installs into /usr/share/cmake (etc.) too.

AndrewGaspar:
> Why is it installing to there, anyways? Any other cmake plugin installs into /usr/share/cmake (etc.) too.

CMake recognizes it as a universal install path that works on both Unix and Windows without having to do any platform-specific behavior. https://cmake.org/cmake/help/latest/command/find_package.html?highlight=find_package#search-procedure

If this package was installing any executables or libraries that would be directly consumed by users, I'd understand your concern, but this package should _only_ be consumed by CMake, so I don't see any reason why it would even need to install its executables directly into $PREFIX/libexec or similar.

This behavior also has the benefit of making the package trivially relocatable, for some marginal benefit.

I can see package managers not liking things to be installed outside `bin`, `lib`, `libexec`, etc. If you make this change a non-default option, I could merge that.

You'll need to update `CorrosionConfig.cmake.in`, too: https://github.com/AndrewGaspar/corrosion/blob/master/cmake/CorrosionConfig.cmake.in

AndrewGaspar:
Made defunct by #78. Thank you for the PR!

PR #74, opened 2020-11-17
Allow turning off compiling the generator

2xsaiko:
For in case it's already installed and you want to use the system install, or just don't want to build it in the first place, e.g. for gentoo packaging which has special handling for Rust projects.


AndrewGaspar:
Could you walk me through workflow a bit more?

My basic understanding of what you want to do - you want to install `corrosion-generator` as one package, and then install the rest of Corrosion (cmake scripts and config files, basically) as a second package?

2xsaiko:
Gentoo ebuilds don't have network access at any time during the actual compile process, so just doing `cargo build` like corrosion does won't work. Instead, it has its own eclass (which is basically a small library that your ebuild can inherit to add e.g. build system specific functions) for Cargo/Rust projects, which I'm using here to build the generator. That part doesn't need any changes in the repo itself, but the cmake project will still try to build the generator too which fails because it can't access the network, so I have to be able to disable that part.

Here's [the ebuild][1], if you're curious.

[1]: https://git.sr.ht/~dblsaiko/ebuilds/tree/master/dev-util/corrosion/corrosion-9999.ebuild

AndrewGaspar:
Could you restructure this PR in this way: Instead of having a `CORROSION_BUILD_GENERATOR` option, instead have a CACHE variable called `CORROSION_GENERATOR_EXECUTABLE` that is unset by default. In `Corrosion.cmake`, in [this branch](https://github.com/AndrewGaspar/corrosion/blob/master/cmake/Corrosion.cmake#L30), check for that variable and, if set, set `_CORROSION_GENERATOR_EXE` to its value. Make sure to document this variable in the README under [Advanced](https://github.com/AndrewGaspar/corrosion#advanced).

If you do that, then I don't think you need the rest of your changes, and you preserve the totality of the install path for Corrosion. From a brief perusal of the source for the cargo eclass, I think that Corrosion should pick up the vendored dependencies created by it.

AndrewGaspar:
Made defunct by #91

PR #75, opened 2020-11-22
Use more standard install directories

JBBgameich:
Use GNUInstallDirs, which should improve compatibility with install layouts across different Linux distributions.
I'm considering to use corrision in a project that is potentially going to be packaged in some distributions, so the install paths need to be standard.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,7 @@ if (NOT CORROSION_IS_SUBDIRECTORY)
 
     corrosion_install(
         TARGETS corrosion-generator
-        DESTINATION Corrosion/libexec
+        DESTINATION libexec
     )
 
     # Generate the Config file
@@ -62,7 +62,7 @@ if (NOT CORROSION_IS_SUBDIRECTORY)
 
     configure_package_config_file(
         cmake/CorrosionConfig.cmake.in CorrosionConfig.cmake
-        INSTALL_DESTINATION Corrosion/lib/cmake/Corrosion
+        INSTALL_DESTINATION lib/cmake/Corrosion
     )
 
     if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.13")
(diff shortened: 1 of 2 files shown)


JBBgameich:
As far as I can see the CI is generally broken right now, not necessarily related to the changes.

AndrewGaspar:
Alright, the sentiment seems to be that the current behavior is undesirable, at least on Unix platforms... It seems I must relent and use the standard install prefix... https://github.com/AndrewGaspar/corrosion/pull/73

cc @2xsaiko

The current behavior is nice because it works for both Unix and Windows. I guess distros don't like things being installed to the GNU standard locations of the install prefix, though. I suppose I can understand that concern

I'll probably end up filing another PR myself to carefully manage the differences between how CMake does package lookup on Windows and Unix.

How would you feel about me placing `corrosion-generator` in a path that is tightly coupled to the `CorrosionConfig.cmake` file? It would be nice to keep the package trivially relocatable. (e.g. `cp -R`)

> As far as I can see the CI is generally broken right now, not necessarily related to the changes.

Ugh, I need to fix this. Filed an issue: https://github.com/AndrewGaspar/corrosion/issues/76

JBBgameich:
I'm not really sure how cmake finds packages on windows, and how the GNUInstallDirs would end up there, but with the current state of this PR, it should still be possible to achieve the install layout you are currently using, simple by setting the CMAKE_INSTALL_PREFIX to `SOME_PATH/Corrosion`

The aim of keeping the package relocatable isn't a bad thing in my opinion, but I'd prefer if Corrosion used the standard directories by default without building up its own prefix somewhere. If it is necessary to build a relocatable package, which is probably not needed in most cases, that should still work by adding a `Corrosion` suffix to the install prefix, similar to what I said about windows.

AndrewGaspar:
> I'm not really sure how cmake finds packages on windows, and how the GNUInstallDirs would end up there

`find_package`  has different default search paths for packages on Windows and Unix: https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure

> it should still be possible to achieve the install layout you are currently using, simple by setting the CMAKE_INSTALL_PREFIX to SOME_PATH/Corrosion

Yes, but the default is broken for Windows with this change, so I'll need to make sure that is set as the default.

> Do you mean "like" instead of "don't like" or do we have some kind of misunderstanding here?

Yes, sorry, got mixed up.

AndrewGaspar:
Alright, addressed the issues in #78 . Thank you for submitting this PR!

Issue #80, opened 2020-12-14
Provide environment variables to learn about CMake properties


humenda:
For the development of procedural macros that interact with the C world, it would be great to expose some of the CMake variables using environment variables. Most prominent would be the build directory. There is already `CMAKECARGO_BUILD_DIR`, but that's only exposed to the CMake config generator, not to the `cargo build` invocation.

humenda:
An alternative would be to allow custom variables to be passed to `corrosion_import_crate`. I would imagine a keyword argument that can occur 0-n times and that allows information such as the build directory, but also any other CMake information to be passed.

AndrewGaspar:
> An alternative would be to allow custom variables to be passed to corrosion_import_crate. I would imagine a keyword argument that can occur 0-n times and that allows information such as the build directory, but also any other CMake information to be passed.

I've actually thought about doing this in the past. For example, let's say you were using Corrosion and an `*-sys` crate in Rust - it would be useful to be able to tell that crate to find libraries, etc. at the same place that CMake found the library.

AndrewGaspar:
I think the [`cmake-file-api`](https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html#v1-client-stateful-query-files) would be helpful here, too.

neachdainn:
> For example, let's say you were using Corrosion and an *-sys crate in Rust - it would be useful to be able to tell that crate to find libraries, etc. at the same place that CMake found the library.

That is doable right now, if a little painful:

```cmake
find_package(foo REQUIRED)
list(TRANSFORM foo_LIBRARIES PREPEND "-L" OUTPUT_VARIABLE RUSTFLAGS)
string(REPLACE ";" " " RUSTFLAGS "${RUSTFLAGS}")
set(ENV{RUSTFLAGS} "${RUSTFLAGS}")
```

After that, you do *not* do any linking inside of Rust. I.e., do *not* have a `#[link(name = "foo")]`.

neachdainn:
I should have double checked before I commented:  something is happening during the build to make the linker not respect `RUSTFLAGS`. Running Cargo directly with the Rust flags (e.g., `RUSTFLAGS="-l/path/to/libfoo.so" cargo run`) works, and passing invalid flags via CMake (e.g., `set(ENV{RUSTFLAGS} "-WAAAGH")`) does cause Cargo to fail, but something is happening during linking that I didn't expect.

neachdainn:
[Found it][1]

[...]

PR #89, opened 2021-02-02
Allow Passing RUSTFLAGS Into Cargo Build

neachdainn:
This change allows the user to specify a set of `RUSTFLAGS` to be passed to Cargo when building by adding a `RUSTFLAGS` argument to `corrosion_import_crate`.

To support this, the generator's `gen_cmake` command was modified to accept an additional `--rustflags` argument that is then written to the generated CMake as a newly added argument to `_add_cargo_build`. When the generator's `build-crate` command is called, the `RUSTFLAGS` environment variable is set to the argument. Naturally, `build-crate` was modified to respect the environment variable.

Based on conversation in #80, I am using this to link to C libraries built/found by CMake. As an example:

```cmake
# CMakeLists.txt
find_package(foo REQUIRED)

set(LINK_DIRS)
foreach(lib IN LISTS foo_LIBRARIES)
    get_filename_component(dir "${lib}" DIRECTORY)
    list(APPEND LINK_DIRS "-L${dir}")
endforeach()
list(REMOVE_DUPLICATES LINK_DIRS)
string(REPLACE ";" " " LINK_DIRS "${LINK_DIRS}")

corrosion_import_crate(MANIFEST_PATH Cargo.toml RUSTFLAGS "${LINK_DIRS}")
```

```rust
// lib.rs
#[link(name = "foo")]
extern "C" { ... }
```

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -65,7 +65,7 @@ set(_CORROSION_RUST_CARGO_TARGET ${Rust_CARGO_TARGET} CACHE INTERNAL "target tri
 
 function(_add_cargo_build)
     set(options "")
-    set(one_value_args PACKAGE TARGET MANIFEST_PATH)
+    set(one_value_args PACKAGE TARGET MANIFEST_PATH RUSTFLAGS)
     set(multi_value_args BYPRODUCTS)
     cmake_parse_arguments(
         ACB
@@ -174,6 +174,7 @@ function(_add_cargo_build)
                 CMAKECARGO_BUILD_DIR=${CMAKE_CURRENT_BINARY_DIR}
                 CMAKECARGO_LINK_LIBRARIES=${link_libs}
                 CMAKECARGO_LINK_DIRECTORIES=${search_dirs}
+                RUSTFLAGS=${ACB_RUSTFLAGS}
                 ${link_prefs}
                 ${compilers}
                 ${lang_targets}
@@ -211,7 +212,7 @@ endfunction(_add_cargo_build)
 
 function(corrosion_import_crate)
     set(OPTIONS)
-    set(ONE_VALUE_KEYWORDS MANIFEST_PATH)
+    set(ONE_VALUE_KEYWORDS MANIFEST_PATH RUSTFLAGS)
     set(MULTI_VALUE_KEYWORDS CRATES)
     cmake_parse_arguments(COR "${OPTIONS}" "${ONE_VALUE_KEYWORDS}" "${MULTI_VALUE_KEYWORDS}" ${ARGN})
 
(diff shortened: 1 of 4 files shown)


neachdainn:
Tests might fail - it seems like there may be something wrong with the dependencies. Looking into it now.

neachdainn:
The issue was just that I had forgotten I had previously forked this repo, so what I assumed was the current master was actually a bit older. Rebasing should have it fixed.

neachdainn:
Pulling the conversation here because I think it is more relevant.

> So instead, I think for now I need to go the other way and just try to tell Cargo as much as possible to specify to the link options. Unfortunately, this is a bit of a bifurcation between static Rust libraries and shared Rust libraries/executables. For the former, the final link step is handled entirely in CMake, which makes some things easier, but in the latter, it's handled entirely in Cargo.

As far as I am concerned for getting this working, that is fine as long as we can identify the appropriate variables to populate. I assumed pulling `RUSTFLAGS` from the environment in the generator would work, but that doesn't seem to be the case.

neachdainn:
@AndrewGaspar This is ready to be looked at and possibly merged. I haven't tested it very thoroughly but it is allowing me to use CMake to link libraries to my Rust crates. It doesn't fully address #80 and it doesn't implement the ideal API you mentioned, though it should enable doing a good number of things until the ideal API is implemented.

AndrewGaspar:
I need to fix something in the CI. GitHub keeps changing crap in the GitHub Actions environments. Looking into that tonight.

Thank you for the PR!

I'm not quite sure this is how I'd like to do this. I think it should be more like [`target_compile_options`](https://cmake.org/cmake/help/latest/command/target_compile_options.html?highlight=target_compile_options). Call it something like `corrosion_compile_options`. That way you can call it multiple times to append additional flags. Do you think you could change this to implement it that way? I think you should be able to do something like `set_property(TARGET cargo-build_${target_name} APPEND PROPERTY CORROSION_RUST_FLAGS ${flags}`, and then apply it in the `add_custom_target` command using a generator expression like `RUSTFLAGS=$<TARGET_PROPERTY:CORROSION_RUST_FLAGS>`. I _think_ that should work. Let me know if you hit difficulties.

neachdainn:
I should be able to find some time to do that this weekend. I think figuring out how to make `corrosion_compile_options` will also put us 90% of the way towards `corrosion_link_libraries` so I might try to get that working as well.

AndrewGaspar:
I did manage to fix the CI, so if you merge master, it should clean up the test failures.

ogoffart:
That PR is another example that uses set_target_property to change stuff https://github.com/AndrewGaspar/corrosion/pull/102
Also, maybe that PR would also allow to specify the RUSTFLAGS env variable, but a explicit way to set the rust flags is probably a better idea.

neachdainn:
In the chaos of life, this pull request totally fell off my radar. I'll try to find time to reconcile this with the comments here and the pull request just linked.

jschwe:
Closing, since https://github.com/corrosion-rs/corrosion/pull/127 added support for Rustflags.

PR #94, opened 2021-03-30
Update cargo_metadata dependency

onContentStop:
Hi! I noticed an outdated dependency. An update shouldn't hurt.


AndrewGaspar:
Thank you!

PR #95, opened 2021-03-31
If MANIFEST_PATH is unwrapped, it should be required.

onContentStop:
Otherwise, the user gets to see a nice panic message instead of
something short and actionable.


AndrewGaspar:
In the current code I don't think there's any code path that can result in `--manfiest-path` not being passed to the generator, but this would be helpful for developers on Corrosion. Looks good!

PR #99, opened 2021-04-12
Added FEATURES option to corrosion_import_crate

### LINKED ISSUE #57: Selectively import crates

kwek20:
Fixed #57 part 2. 

I have not yet managed to solve the features using `set_property`


kwek20:
Apparently the set_property restriction for interfaces is removed in 3.19.x, but still couldnt get it to work though
 `get_target_property() called with non-existent target "rust-lib"`
 
 Any ideas @AndrewGaspar ?

ogoffart:
I think being able to set the features is indeed needed. I'm probably going to make use of that at some point.

I'm just afraid that the way this is implemented in this patch doesn't work well with workspace manifest where one select only a crate, and want to enable features for that one crate.

tronical:
What's the restriction you're running into with `set_property`?

In #102 I have a patch for review that kind of works similar to that, allowing the user to set a custom property on the targets created by `corrosion_import_crate` which is passed on to the cargo build.

If there was a `CORROSION_FEATURES` property, then perhaps it could be added to the cargo command line by passing the feature list to the generator using `$<TARGET_PROPERTY:${target_name},CORROSION_FEATURES>` ?

kwek20:
Hi guys! Thanks for the feedback, I do not recall why it didnt work, im pretty new to cmake so it could just be i did it wrong .)
I wont have time to fix it in the coming 3 weeks, but after that I can give it another shot

tronical:
Hi! I've taken @kwek20 's initial commit and changed it to use a target property and submitted it as #108

kwek20:
Oooh! Ill give it a test! Thanks man.. you know how life gets in the way of things :) @tronical

tronical:
> Oooh! Ill give it a test! Thanks man.. you know how life gets in the way of things :) @tronical

Haha yeah, I know that feeling :)

PR #100, opened 2021-04-23
Use `USES_TERMINAL` instead of job pools

ogoffart:
This will make sure that ninja does not buffer all the output from cargo
so that the user can see cargo progress in the terminal as the build
proceeds.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -24,8 +24,6 @@ get_property(
     TARGET Rust::Cargo PROPERTY IMPORTED_LOCATION
 )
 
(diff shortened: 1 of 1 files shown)


ogoffart:
@AndrewGaspar  could you please merge?

I use JOB_POOL in #68  , but actually, the USES_TERMINAL is much better.  USES_TERMINAL implies a job pool as it will be in the one terminal pool which can only have one job at the time, so there will not be a "Blocking waiting for file lock on build directory" message either. But what's better is that the output of cargo will be displayed to the terminal immediately instead of being buferred. Because running cargo can take so long, and cargo show output in color and with progress, this is a great improvement.

PR #102, opened 2021-05-04
Add support for specifying environment variables during cargo build

tronical:


test/envvar/build.rs
```diff
@@ -0,0 +1,9 @@
+fn main() {
+    let value = env!("REQUIRED_VARIABLE");
+    if value != "EXPECTED_VALUE" {
+        panic!("unexpected value for required environment variable");
+    }
+    if env!("ANOTHER_VARIABLE") != "ANOTHER_VALUE" {
```
ogoffart, line 6: maybe check with std::env::var.

I suggest this:

```rust
assert_eq!(env!("REQUIRED_VARIABLE"), "EXPECTED_VALUE");
assert_eq!(std::env::var("ANOTHER_VARIABLE").unwrap(),  "ANOTHER_VALUE");
```

tronical: Sure, done

ogoffart:
Looks good to me

tronical:
Note to self: Guard this new feature with cmake version checks

tronical:
I added CMake 3.19 version guards around the tests and documented this in the README. Please approve a CI test run :-)

tronical:
Even querying the properties would cause an error with the old cmake version - even if not set :(. So I added a guard and tried that also with a local cmake version.

(As a side note and unrelated to my tests: On my 11.3 system with Xcode 12.2 the tests don't compile due to some -lSystem linkage issue. The test I added however does compile)

PR #105, opened 2021-06-22
cross compile cdylib fixes

tronical:
This merge request includes a few fixes for cross-compiling when creating cdylibs with Rust as well as when creating host-tools with Rust and using them from CMake.

Reviewing the individual commits might be easiest since the commit messages try to elaborate a bit where applicable. If desired I can also split up this merge requests into multiple ones.

Test coverage is provided for the linker use fixes as well as the host program, not for the CMAKE_SYSROOT passing. Ideas for how to accomplish that would be appreciated, otherwise maybe this one will have to live without a test :)


test/hostbuild/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+corrosion_import_crate(MANIFEST_PATH ${CMAKE_CURRENT_SOURCE_DIR}/Cargo.toml)
+
+set_property(
+    TARGET rust-host-program
+    PROPERTY CORROSION_USE_HOST_BUILD 1
+)
```
ogoffart, line 6: Maybe this test could actually run the rust-host-program from cmake from a custom command

ogoffart (approved):

test/hostbuild/CMakeLists.txt
tronical: I can do that, although I think that ends up testing more general cmake features than corrosion. How strong do you feel about it? :)

tronical:
Thanks for the review and merge :)

PR #106, opened 2021-06-28
Fix unnecessary rebuilds when using CORROSION_USE_HOST_BUILD

tronical:
Commit 3c9ff60f51cac21fcd4add47c1a14da47fe1c02e introduced the feature
that host binaries can be built, which is done by omitting the
`--target` parameter to cargo. This however implies that the target host
directory is picked by cargo (for example `cargo/build/debug`) for these
builds, which ends up being shared with the corrosion build itself, as
that is built implicitly via `cargo run`. If there were any common
crates as dependencies, then those may end up being built with different
flags (for corrosion and for host builds of the project), resulting in
unnecessary rebuilds.

To fix this, this patch makes use of the internal knowledge about the
host target by always passing it again.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -60,6 +60,7 @@ set(
 
 set(_CORROSION_CARGO_VERSION ${Rust_CARGO_VERSION} CACHE INTERNAL "cargo version used by corrosion")
 set(_CORROSION_RUST_CARGO_TARGET ${Rust_CARGO_TARGET} CACHE INTERNAL "target triple used by corrosion")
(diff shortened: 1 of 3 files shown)


ogoffart:
> If there were any common crates as dependencies, then those may end up being built with different flags 

Do you know what are the flags differences?  It is annoying that cargo doesn't put the artefact with different cache key in the target directory, so the artifacts could be reused if there were using the same flags.

Another alternative would be to explicitly use the `--target-dir` flag.

But I guess this patch is actually fine as well.

tronical:
> Do you know what are the flags differences? It is annoying that cargo doesn't put the artefact with different cache key in the target directory, so the artifacts could be reused if there were using the same flags.

For example if the host tool depends on `clap` then that's shared with `corrision-generator`'s clap dependency, which uses for example `default-features = false`.

PR #107, opened 2021-08-29
Allow using FindRust.cmake several times

gladhorn:
CMake only allows creating alias targets once, otherwise it bails out with
CMake Error at thirdparty/corrosion/cmake/FindRust.cmake:330 (add_executable):
  add_executable cannot create imported target "Rust::Rustc" because another
  target with the same name already exists.

Allow using find_package(Rust) in several places by following the pattern of other
find modules, protecting with an if.

### DIFF
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -340,14 +340,16 @@ else()
     _gen_config(Debug OFF)
 endif()
 
-add_executable(Rust::Rustc IMPORTED GLOBAL)
-set_property(
-    TARGET Rust::Rustc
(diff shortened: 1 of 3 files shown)


ogoffart:
Hi  Frederik 😉,
Looks good to me.
A test would be nice.

ogoffart:
Hi  Frederik 😉,
Looks good to me.
A test would be nice.

gladhorn:
Hello Olivier 😁 , test added.

ogoffart:
Thank you :+1:

PR #108, opened 2021-09-02
Add support for feature selection

tronical:
This merge requests is based on the commit in  #99 but changes the code to use a target property.


README.md
```diff
@@ -152,6 +152,10 @@ The pairs can also be generator expressions, as they will be evaluated using $<G
 For example this may be useful if the build scripts of crates look for environment variables.
 This feature requires CMake >= 3.19.0.
 
+Cargo crates sometimes offer features, which traditionally can be specified with `cargo build` on the
+command line as opt-in. You can select the features to use for a crate imported with Corrosion by
+setting the `CORROSION_FEATURES` target list property on the targets created by `corrosion_import_crate`.
```
ogoffart, line 157: Maybe the README would benefit from an example

ogoffart (commented):
Looks good.

I think it might still make sense to be able to do 
```cmake
corrosion_import_crate(MANIFEST_PATH rust/Cargo.toml FEATURES myfeature)
```
in the common case, don't you think?

tronical:
> Looks good.
> 
> I think it might still make sense to be able to do
> 
> ```cmake
> corrosion_import_crate(MANIFEST_PATH rust/Cargo.toml FEATURES myfeature)
> ```
> 
> in the common case, don't you think?

Hm, that's true. So `FEATURES` will apply to all the imported crates equally?

ogoffart:
> Hm, that's true. So FEATURES will apply to all the imported crates equally?

yeah

cmake/Corrosion.cmake
```diff
@@ -171,8 +171,24 @@ function(_add_cargo_build)
...
+
     if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.19.0)
         set(build_env_variable_genex "$<GENEX_EVAL:$<TARGET_PROPERTY:${target_name},CORROSION_ENVIRONMENT_VARIABLES>>")
+        
+        set(features_target_property "$<GENEX_EVAL:$<TARGET_PROPERTY:${target_name},CORROSION_FEATURES>>")
+        set(features_genex "$<$<BOOL:${features_target_property}>:--features=$<JOIN:${features_target_property},$<COMMA>>>")
+
+        # target property overrides corrosion_import_crate argument
+        set(all_features_target_property "$<GENEX_EVAL:$<TARGET_PROPERTY:${target_name},CORROSION_ALL_FEATURES>>")
+        set(all_features_arg "$<IF:$<BOOL:${all_features_target_property}>,${all_features_target_property},${all_features_arg}>")
```
tronical, line 189: This override isn’t working, it’s missing the Boolean to argument transformation.

tronical: This is fixed now

ogoffart (approved):

PR #111, opened 2021-09-14
make FindRust gracefully fail if rustc not found

trondhe:
Motivation for this change is to be able to reuse the FindRust module in a more general setting to see if rust exists.
With this change, it should be able to be used by any other package and afterwards check Rust_FOUND of its existence.

Earlier a missing .cargo/bin/rustc would fail with
```
The system cannot find the file specified
CMake Error at corrosion/cmake/FindRust.cmake:171 (get_filename_component):
  get_filename_component called with incorrect number of arguments
```
Now it prints a more descriptive errormessage, sets Rust_FOUND to false
and returns.
It also sets Rust_FOUND to true if a rust target is successfully created

### DIFF
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -166,9 +166,19 @@ if (_RESOLVE_RUSTUP_TOOLCHAINS)
             NO_DEFAULT_PATH)
 else()
     find_program(Rust_COMPILER_CACHED rustc)
+    if (Rust_COMPILER_CACHED)
+        get_filename_component(_RUST_TOOLCHAIN_PATH ${Rust_COMPILER_CACHED} DIRECTORY)
+        get_filename_component(_RUST_TOOLCHAIN_PATH ${_RUST_TOOLCHAIN_PATH} DIRECTORY)
+    endif()
+endif()
 
-    get_filename_component(_RUST_TOOLCHAIN_PATH ${Rust_COMPILER_CACHED} DIRECTORY)
-    get_filename_component(_RUST_TOOLCHAIN_PATH ${_RUST_TOOLCHAIN_PATH} DIRECTORY)
+if (NOT Rust_COMPILER_CACHED)
+    message(
+        WARNING "The rustc executable was not found. "
+        "Rust not installed or ~/.cargo/bin not added to path? "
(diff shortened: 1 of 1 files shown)


cmake/FindRust.cmake
```diff
@@ -166,9 +166,19 @@ if (_RESOLVE_RUSTUP_TOOLCHAINS)
...
+        get_filename_component(_RUST_TOOLCHAIN_PATH ${_RUST_TOOLCHAIN_PATH} DIRECTORY)
+    endif()
+endif()
 
-    get_filename_component(_RUST_TOOLCHAIN_PATH ${Rust_COMPILER_CACHED} DIRECTORY)
-    get_filename_component(_RUST_TOOLCHAIN_PATH ${_RUST_TOOLCHAIN_PATH} DIRECTORY)
+if (NOT Rust_COMPILER_CACHED)
+    message(
+        WARNING "The rustc executable was not found. "
+        "Rust not installed or ~/.cargo/bin not added to path? "
```
ratijas, line 178: > ~/.cargo/bin not added to path?

That should be trivial to check automatically, and print even more descriptive message.

ratijas (commented):
Looks cool. Just one little comment there.

cmake/FindRust.cmake
trondhe: Yeah, I am not sure what the best actual check would be for doing this.
I did this PR as a more minimal improvement, but there is likely a better check which can be done alongside a refactor to make it more clear.
You have any suggestions? Because the .cargo/bin could probably be placed in different paths, so it needs to support just searching for the rustc executable in path I guess.

trondhe: @ratijas You think this is a fine first step, or do you want me to do any larger changes?

ratijas: Ship it.

Even if wording is not 100% technically correct and ommits some non-standard setups, it still nudges user in the right direction.

trondhe:
@ratijas I have no write access, so I can't merge :)

ogoffart:
Thanks

ratijas:
Me neither. I'm a contributor, not collaborator — although was almost promoted to be one. I think I still I lack CMake skill a bit, but gradually getting there :)

PR #112, opened 2021-09-14
Relax Cargo version parsing

jschwe:
Some cargo versions shipped with rust nightlies seem to be reporting their version without the cargo prefix, e.g.:
```
$ cargo +nightly-2021-07-24 --version
1.55.0-nightly (cebef2951 2021-07-22)
```

 This commit relaxes the parsing slightly by adding an elseif
 condition that matches on the expression without the leading
 `cargo `. It's not an elegant solution, but it works for now.

______________

This is just a quick workaround from my side, but I'm open to suggestions if you would like a more thorough fix.

### DIFF
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -198,6 +198,12 @@ if (_CARGO_VERSION_RAW MATCHES "cargo ([0-9]+)\\.([0-9]+)\\.([0-9]+)")
     set(Rust_CARGO_VERSION_MINOR "${CMAKE_MATCH_2}")
     set(Rust_CARGO_VERSION_PATCH "${CMAKE_MATCH_3}")
     set(Rust_CARGO_VERSION "${Rust_CARGO_VERSION_MAJOR}.${Rust_CARGO_VERSION_MINOR}.${Rust_CARGO_VERSION_PATCH}")
+# Workaround for the version strings where the `cargo ` prefix is missing.
+elseif(_CARGO_VERSION_RAW MATCHES "([0-9]+)\\.([0-9]+)\\.([0-9]+)")
+    set(Rust_CARGO_VERSION_MAJOR "${CMAKE_MATCH_1}")
+    set(Rust_CARGO_VERSION_MINOR "${CMAKE_MATCH_2}")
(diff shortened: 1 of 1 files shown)


cmake/FindRust.cmake
```diff
@@ -198,6 +198,12 @@ if (_CARGO_VERSION_RAW MATCHES "cargo ([0-9]+)\\.([0-9]+)\\.([0-9]+)")
     set(Rust_CARGO_VERSION_MINOR "${CMAKE_MATCH_2}")
     set(Rust_CARGO_VERSION_PATCH "${CMAKE_MATCH_3}")
     set(Rust_CARGO_VERSION "${Rust_CARGO_VERSION_MAJOR}.${Rust_CARGO_VERSION_MINOR}.${Rust_CARGO_VERSION_PATCH}")
+# Workaround for the version strings where the `cargo ` prefix is missing.
+elseif(_CARGO_VERSION_RAW MATCHES "([0-9]+)\\.([0-9]+)\\.([0-9]+)")
```
ratijas, line 202: Can't we just use optional (zero-or-one) quantifier though? Like this: `"(cargo )?([0-9…`

jschwe: That would definitely be nicer. However, it seems that CMake does not support non-capturing groups. At least the [regex specification](https://cmake.org/cmake/help/latest/command/string.html#regex-specification) does not mention it, and configuring corrosion fails because of the unexpected `?` after the opening `(`:

```
cmake -S. -Bbuild -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$HOME/.local/
-- Rust Toolchain: stable-x86_64-unknown-linux-gnu
RegularExpression::compile(): ?+* follows nothing.
RegularExpression::compile(): Error in compile.
CMake Error at cmake/FindRust.cmake:197 (if):
  if given arguments:

    "_CARGO_VERSION_RAW" "MATCHES" "(?:cargo )?([0-9]+)\\.([0-9]+)\\.([0-9]+)"
``` 


I guess I could do if ( a <OR> b) to remove the code duplication of the `set()` commands.

ratijas: That's pretty bad. I'm gonna file a bug upstream.

ogoffart:
Thanks!

PR #118, opened 2021-12-06
Export flag to check if rustc is a nightly

jschwe:
Since using unstable features (typically) requires a nightly toolchain,
check by matching against the `-nightly` suffix of the rustc version and
export that information as a new flag `Rust_IS_NIGHTLY` to users.
A user could then conditionally enable unstable features, by selecting a
corresponding feature via corrosion.

It seems that currently variables such as `Rust_VERSION` are not documented, so I also did not add documentation for this new flag. Would you accept an MR that documents the `Rust_` variables that `FindRust.cmake` provides?

### DIFF
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -224,11 +224,14 @@ execute_process(
     COMMAND ${Rust_COMPILER_CACHED} --version --verbose
     OUTPUT_VARIABLE _RUSTC_VERSION_RAW)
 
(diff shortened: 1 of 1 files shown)


ogoffart:
Looks alright, thank you.

The CI failure must be unrelated.

ogoffart (approved):
But i don't know how to merge if the CI is failling

ogoffart:
I'm merging this since the CI faillure was unrelated.

PR #120, opened 2021-12-18
Attempt to fix CI

jschwe:
- Fix Building on Windows with Rust >= 1.57 by linking with Bcrypt.
- Test with MacOS 10.15 instead of `macos-latest`, since linking on MacOS 11 currently does not work. I don't have a mac, so I'm leaving it up to someone who has one to fix building on Macos 11.

#Closes 119

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -20,7 +20,7 @@ jobs:
         os:
           - windows-latest
           - ubuntu-latest
-          - macos-latest
+          - macos-10.15 # macos-11 is currently not working.
         arch:
           - x86_64
           - i686
@@ -69,11 +69,11 @@ jobs:
             generator: ninja
           - os: windows-latest
             arch: powerpc64le
-          - os: macos-latest
+          - os: macos-10.15
             arch: i686
-          - os: macos-latest
+          - os: macos-10.15
             arch: aarch64
-          - os: macos-latest
+          - os: macos-10.15
             arch: powerpc64le
 
           # ABI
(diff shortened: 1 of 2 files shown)


ogoffart:
thanks a lot.

ogoffart:
i don't understand why the CI is still yellow.  Looks like a github bug.

ogoffart:
Closing/re-opening restarted the CI. Let's see what happens.

ogoffart:
It looks like there if still one job using macos-latest that does not work.
Also the matrix is a bit big and we probably should reduce it.

jschwe:
> It looks like there if still one job using macos-latest that does not work. Also the matrix is a bit big and we probably should reduce it.

Could you go to "Settings" (of this repository) and then to "Branches" / "Require status checks to pass before merging" and remove the `macos-latest` jobs from that list?

In my fork it looks something like this:
grafik

ogoffart:
I don't have the rights to do that.
@AndrewGaspar , can you please do it?

jschwe:
@AndrewGaspar Could you also do this for the remaining two required jobs:
- Test Corrosion as a Library (macos-latest, install) 
- Test Corrosion as a Library (macos-latest, subdirectory)

On a related note: It seems that the failing `macos-latest` CI tests do work if you use CMake `3.19`. I don't have a Mac so I can't really say anything more than that.

ogoffart:
I don't even know where the ` Test Corrosion as a Library (macos-latest, install)` and `Test Corrosion as a Library (macos-latest, subdirectory)` tests comes from. It's not in the github action.
I now have the privilege to merge this so i'll do that.
Thanks for your help.

PR #121, opened 2021-12-30
Bump dependencies 

jschwe:
- Bumps the dependencies of the Generator to the most recent versions
  - `platforms` renamed some enums
- Since bitflags (dependency of clap) version 1.3 raises the MSRV to 1.46, specify bitflags as a direct dependency to prevent `cargo update` from updating bitflags beyond `1.2`.  `clap` specifies the required bitflags versions as `^1`.

### DIFF
--- a/generator/Cargo.toml
+++ b/generator/Cargo.toml
@@ -6,12 +6,12 @@ license = "MIT"
 edition = "2018"
 
 [dependencies]
-cargo_metadata = "0.13"
-semver = "0.11"
-platforms = "1.1"
+cargo_metadata = "0.14"
+semver = "1.0"
+platforms = "2.0"
 
 [dependencies.clap]
-version = "2.33"
+version = "2.34"
 default-features = false
 
 # Make sure this crate still compiles while it is checked out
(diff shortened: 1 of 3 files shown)


generator/Cargo.toml
```diff
@@ -6,12 +6,16 @@ license = "MIT"
...
-cargo_metadata = "0.13"
-semver = "0.11"
-platforms = "1.1"
+cargo_metadata = "0.14"
+semver = "1.0"
+platforms = "2.0"
+# Bitflags is pulled in by clap, but 1.3 bumps the MSRV to 1.46,
+# so we ensure cargo doesn't update beyond 1.2 by specifying it 
+# as a direct dependency.
+bitflags = "~1.2.1"
```
ogoffart, line 15: I don't think we should do that.
We should bump our own MSRV, or keep the old version in the Cargo.lock

ogoffart (changes requested):

generator/Cargo.toml
jschwe: Okay, I dropped that commit.

ogoffart:
Thanks

Issue #116, opened 2021-11-11
Option CORROSION_INSTALL_EXECUTABLE not working 


schultzern:
Hello!

First of all thank you for the work on this repo :star:. 

I'm having issues using the option `CORROSION_INSTALL_EXECUTABLE` documented [here](https://github.com/AndrewGaspar/corrosion#developermaintainer-options). 

```
git clone https://github.com/AndrewGaspar/corrosion.git
...
cd corrosion
cmake -S./ -B./build3 -DCMAKE_BUILD_TYPE=Release -DCORROSION_INSTALL_EXECUTABLE=0 -DCMAKE_INSTALL_PREFIX="${PWD}/build3/install"
-- The CXX compiler identification is GNU 11.1.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Rust Target: x86_64-unknown-linux-gnu
-- Found Rust: /usr/bin/rustc (found version "1.56.0") 
-- Using Corrosion as a subdirectory
-- Configuring done
-- Generating done
-- Build files have been written to: /home/esc/ss/rust/corrosion/repo/build3
esc@esc-home ~/.../corrosion/repo (master)$ cmake --build build3 --config Release
   Compiling proc-macro2 v1.0.26
   Compiling unicode-xid v0.2.1
   Compiling syn v1.0.71
   Compiling serde_derive v1.0.125
   Compiling serde v1.0.125
   Compiling ryu v1.0.5
   Compiling ucd-trie v0.1.3
   Compiling bitflags v1.2.1
   Compiling serde_json v1.0.64
   Compiling camino v1.0.4
   Compiling platforms v1.1.0
   Compiling unicode-width v0.1.8
   Compiling itoa v0.4.7
   Compiling textwrap v0.11.0
   Compiling pest v2.1.3
   Compiling clap v2.33.3
   Compiling quote v1.0.9
   Compiling semver-parser v0.10.2
   Compiling cargo-platform v0.1.1
   Compiling semver v0.11.0
   Compiling cargo_metadata v0.13.1
   Compiling corrosion-generator v0.1.0 (/home/esc/ss/rust/corrosion/repo/generator)
    Finished release [optimized] target(s) in 12.48s
[  0%] Built target cargo-build_corrosion-generator
[...]

Issue #84, opened 2021-01-19
CorrosionConfig.cmake looks in wrong place for Corrosion::Generator when CMAKE_INSTALL_LIBEXEC is an absolute path


surprized:
When packaging for openSUSE, we use the RPM macro ```%cmake``` to set default install directories, so that the command expands to:
[...]

PR #122, opened 2021-12-31
Fix CORROSION_INSTALL_EXECUTABLE option and respect absolute paths for CMAKE_INSTALL_LIBEXECDIR

jschwe:
- Commit Nr. 1 should fix https://github.com/corrosion-rs/corrosion/issues/116.
- Commit Nr.2 aims to fix https://github.com/corrosion-rs/corrosion/issues/84. I'm not really sure what @AndrewGaspar had in mind, but I think this should work as a band-aid sort of fix for the issue.


ogoffart:
Thanks!

For the relative path stuff, I think this should be done with `file(REAL_PATH` but since this was only added in recent version of cmake, i guess what you did is the correct way.


I still don't understand why gihub things there are CI jobs in macos-latest

jschwe:
> Thanks!
> 
> For the relative path stuff, I think this should be done with `file(REAL_PATH` but since this was only added in recent version of cmake, i guess what you did is the correct way.

Thanks, I wasn't aware of `file(REAL_PATH)`. But I guess in this case it doesn't really matter, as you said.


> I still don't understand why gihub things there are CI jobs in macos-latest

That's probably because those two jobs are still selected as "Required" to pass in the settings of the github repository. Andrew already removed most of them, but probably missed the remaining two. I'll open an issue and tag him, so it should get resolved eventually.

ogoffart:
> That's probably because those two jobs are still selected as "Required" to pass in the settings of the github repository. Andrew already removed most of them, but probably missed the remaining two. I'll open an issue and tag him, so it should get resolved eventually.

Ah yes, i removed them.

PR #123, opened 2022-01-07
Improve documentation of Rust version related variables

jschwe:
- Document the version related variables that corrosion (FindRust) provides
- Improve the Rust_IS_NIGHTLY variable by always defining it (allows testing wether corrosion is new enough to set this variable)
- Add variables to expose the LLVM version used by rust. Useful e.g. for cross-language lto, where you want to be sure that the llvm version on both sides matches.

### DIFF
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -108,7 +108,7 @@ if (_RESOLVE_RUSTUP_TOOLCHAINS)
                 set(_TOOLCHAIN_OVERRIDE ${_TOOLCHAIN})
             endif()
         else()
-            message(WARNING "Didn't reconize toolchain: ${_TOOLCHAIN_RAW}")
+            message(WARNING "Didn't recognize toolchain: ${_TOOLCHAIN_RAW}")
         endif()
     endforeach()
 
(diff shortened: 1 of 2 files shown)


cmake/FindRust.cmake
```diff
@@ -231,6 +231,8 @@ if (_RUSTC_VERSION_RAW MATCHES "rustc ([0-9]+)\\.([0-9]+)\\.([0-9]+)(-nightly)?"
     set(Rust_VERSION "${Rust_VERSION_MAJOR}.${Rust_VERSION_MINOR}.${Rust_VERSION_PATCH}")
     if(CMAKE_MATCH_4)
         set(Rust_IS_NIGHTLY 1)
+    else()
+        set(Rust_IS_NIGHTLY 0)
```
ogoffart, line 235: Will that cause problem for people doing
`if(Rust_IS_NIGHTLY)`  or is 0 falsy?

cmake/FindRust.cmake
```diff
@@ -248,6 +250,18 @@ else()
     )
 endif()
 
+if (_RUSTC_VERSION_RAW MATCHES "LLVM version: ([0-9]+)\\.([0-9]+)\\.([0-9]+)")
```
ogoffart, line 253: do users need to care about the LLVM version.
I guess you added that because you have an usecase.

cmake/FindRust.cmake
jschwe: Yes, 0 evaluates to false. From the [documentation](https://cmake.org/cmake/help/latest/command/if.html#basic-expressions)
>  False if the constant is 0, OFF, NO, FALSE, N, IGNORE, NOTFOUND, the empty string, or ends in the suffix -NOTFOUND

cmake/FindRust.cmake
jschwe: My usecase is cross-language LTO. It requires that the LLVM version on both sides matches, so I need to know the version of LLVM used by rust.

jschwe:
@ogoffart Are there still any open points from you, or could I go ahead and merge this?

ogoffart:
No more open points.  Thanks for the patch

PR #124, opened 2022-01-07
Release corrosion v0.1.0

jschwe:
Corrosions version has never been changed yet. Still, new features are
being added, and it would be great to update the version number from
time to time so we can use `find_package` to require a minimal version
number if we need some new corrosion feature.

Therefor I propose to bump the corrosion version number to 0.1.0 and
bump the patch version regularly if new features / bugfixes are added,
and bump the minor for breaking changes, including an increase of the
minimum supported cmake or rust version.

I'm aware that there are still open issues in the Project Plan for v0.1, but I think that this shouldn't keep us from releasing a new version from time to time. 

Any opinions on this?

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,0 +1,31 @@
+# 0.1.0 (2022-02-01)
+
+This is the first release of corrosion after it was moved to the new corrosion-rs organization.
+Since there are no previous releases, this is not a complete changelog but only lists changes since
+September 2021.
+
+## New features
+- [Add --profile support for rust >= 1.57](https://github.com/corrosion-rs/corrosion/pull/130):
+  Allows users to specify a custom cargo profile with 
+  `corrosion_import_crate(... PROFILE <profilename>)`.
+- [Add support for specifying per-target Rustflags](https://github.com/corrosion-rs/corrosion/pull/127):
+  Rustflags can be added via `corrosion_add_target_rustflags(<target_name> [rustflags1...])`
+- [Add `Rust_IS_NIGHTLY` and `Rust_LLVM_VERSION` variables](https://github.com/corrosion-rs/corrosion/pull/123):
+  This may be useful if you want to conditionally enabled features when using a nightly toolchain
+  or a specific LLVM Version.
+- [Let `FindRust` fail gracefully if rustc is not found](https://github.com/corrosion-rs/corrosion/pull/111):
+  This allows using `FindRust` in a more general setting (without corrosion).
+- [Add support for cargo feature selection](https://github.com/corrosion-rs/corrosion/pull/108):
+  See the [README](https://github.com/corrosion-rs/corrosion#cargo-feature-selection) for details on
+  how to select features.
+
+
+## Fixes
+- [Fix the cargo-clean target](https://github.com/corrosion-rs/corrosion/pull/129)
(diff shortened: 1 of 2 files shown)


ogoffart:
Yeah, make sense.
We also probably should have a tag or a branch so we can point to that in `FetchContent_Declare`

jschwe:
Yes, it definitely makes sense to push a tag when we bump the version, and probably also to do a github release and write a short description of the release.

jschwe:
@ogoffart I wrote up some Release notes listing changes since September 2021 (arbitrarily chosen). Do you want any suggestions / additions / changes or can I go ahead and merge this and then create a release?

RELEASES.md
```diff
@@ -0,0 +1,31 @@
+# 0.1.0 (2022-TODO)
```
jschwe, line 1: Todo: replace with a real date before merging

ogoffart:
Thanks for your work!
And thank for creating a ChangeLog 

What i'm thinking is to change the README to recommend something like

```
FetchContent_Declare(
    Corrosion
    GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git
    GIT_TAG origin/stable-0.1
)
```

Or something like that. where `stable-0.1` is a branch that we update everytime we release a new version `v0.1.x` of corrosion (in a semver compatible way)
Name of the branch to be bikeshed.

But this should be done after the release.

So the way I see it, to proceed, would be something like
 1. merge this (after updating the date in the changelog)
 2. do a `v0.1.0` tag and a `stable-0.1`
 3. change the README
 
Can we wait until Monday before merging? I'd like to do some double checking before.

jschwe:
> Can we wait until Monday before merging? I'd like to do some double checking before.

Sure, no rush from my side.

>What i'm thinking is to change the README to recommend something like [...]

I'm honestly not a huge fan of pointing to a branch. I think updating should be explicit since it is always possible that some seemingly harmless fix causes a new regression. The CMake [documentation](https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_declare) even recommends explicitly using the full commit hash instead of a git tag or branch name. Of course, people are free to use a branch name (and I'm not against providing `stable-0.x` branches and mentioning they exist), but I don't think we should recommend using a branch over a tag.


> [...]in a semver compatible way

This may be a bit pedantic, but just to make sure we are talking about the same thing - The official semver 2.0 guarantees for versions below 1:

> [4. Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.](https://semver.org/)

I assume you probably mean something like the cargo semver semantics:
> [Initial development releases starting with "0.y.z" can treat changes in "y" as a major release, and "z" as a minor release. "0.0.z" releases are always major changes. This is because Cargo uses the convention that only changes in the left-most non-zero component are considered incompatible.](https://doc.rust-lang.org/cargo/reference/semver.html)

I think it would make sense to follow the cargo semver semantics here, but we should add that to the readme then.

ogoffart:
Ok, let's go ahead with this.

> I'm honestly not a huge fan of pointing to a branch. I think updating should be explicit since it is always possible that some seemingly harmless fix causes a new regression.

You're the second person I talked to with this opinion. So be it.

(And yes, i consider the same semver definition as Cargo where 0.y.z is treated like y.z)

jschwe:
> Ok, let's go ahead with this.

Okay, I edited the date in `RELEASES.md` and will merge and create a release after CI finishes.

ogoffart:
:partying_face:

PR #126, opened 2022-01-08
Fix corrosion_set_linker_language

crabtw:
It seems that LINKER_LANGUAGE of custom target in generator expression
is always empty so make it as custom property.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -146,7 +146,7 @@ function(_add_cargo_build)
         set(build_type_dir $<IF:$<OR:$<CONFIG:Debug>,$<CONFIG:>>,debug,release>)
     endif()
 
-    set(target_linker_language "$<TARGET_PROPERTY:cargo-build_${target_name},LINKER_LANGUAGE>")
+    set(target_linker_language "$<TARGET_PROPERTY:cargo-build_${target_name},CARGO_LINKER_LANGUAGE>")
 
     # When cross-compiling, fall back to specifying at least the cross-compiling C linker, unless
     # an override is set.
@@ -362,7 +362,7 @@ endfunction(add_crate)
 function(corrosion_set_linker_language target_name language)
     set_property(
         TARGET cargo-build_${target_name}
-        PROPERTY LINKER_LANGUAGE ${language}
+        PROPERTY CARGO_LINKER_LANGUAGE ${language}
     )
 endfunction()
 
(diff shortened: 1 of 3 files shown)


ogoffart:
I don't understand what this fixes.
Is it possible to make a test?

crabtw:
> I don't understand what this fixes. Is it possible to make a test?

Added a test in `envvar`

crabtw:
Actually my project don't use corrosion_set_linker_language. I was trying to figure out how corrosion's LINKER_LANGUAGE works and found that the value is always empty.

jschwe:
> I suspect however that the original idea of using this property was to respect's CMake's choice of linker (via CMAKE__LINKER_PREFERENCE), which is something that goes away with this workaround of prefixing.

We do still have the loop in build_crates.rs which chooses the language with the highest CMAKE_<lang>_LINKER_PREFERENCE, so I think prefixing would be ok.

I do have to admit that `LINKER_LANGUAGE` confuses me. There are the (probably related) genexes `$<LINK_LANGUAGE>, which have [some constraints](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#constraints-link-language-generator-expression). So I think prefixing the override option and if not set choosing the language with the highest linker preference ourselves may indeed be the better option.

jschwe:
> I was trying to figure out how corrosion's LINKER_LANGUAGE works and found that the value is always empty.

I can confirm that.

If `tgt` is a rust crate (so a CMake interface library), this produces an empty file:
```
set_target_properties(${tgt} PROPERTIES LINKER_LANGUAGE C)
file(GENERATE OUTPUT ${tgt}_linker_language.txt CONTENT
       "$<TARGET_PROPERTY:${tgt},LINKER_LANGUAGE>")
```

And this produces a file with "C".
```
set_target_properties(${tgt} PROPERTIES AA_LINKER_LANGUAGE C)
file(GENERATE OUTPUT ${tgt}_linker_language.txt CONTENT
       "$<TARGET_PROPERTY:${tgt},AA_LINKER_LANGUAGE>")
```

So I do think this makes sense.

jschwe:
@ogoffart Since we have released version `0.1`, I would go ahead and merge this and the other open PR if that is okay with you.

ogoffart:
Ok.  I must admit i don't understand this one. But if you do, go ahead!

jschwe:
I had another look at this, and openend issue #133 for a more general discussion of what we should do.

jschwe:
Closing, since #208 deprecated corrosion_set_linker_language.

PR #127, opened 2022-01-17
Add support for RUSTFLAGS

jschwe:
Since https://github.com/corrosion-rs/corrosion/pull/89 has not seen any activity in a while, I took the liberty to implement this while considering the review comments on the original MR. 
I also noticed that some of our newly added features like support for setting "FEATURES", should be possible for CMake < 3.19, as long as we prefix the property name for those versions with `INTERFACE_` (at least for CMake < 3.19).  If there are no objections, I would do a follow-up MR to support this.

This allows RUSTFLAGS to be set via the new INTERFACE_CORROSION_RUSTFLAGS property on a CMake target. The INTERFACE prefix is necessary for CMake versions before 3.19. To hide the target property name, we also expose corrosion_add_target_rustflag() as a helper function to add rustflags to a target.
 Special characters such as double quotes in Rustflags need to be escaped (at least if we want to support windows/Visual studio generator).


generator/src/subcommands/build_crate.rs
```diff
@@ -92,6 +101,8 @@ pub fn invoke(
         cargo.arg("--release");
     }
 
+    let mut rustflags = vec![matches.value_of(RUSTFLAGS).unwrap_or_default().to_owned()];
```
ogoffart, line 104: Don't this create a vector with an empty string instead of an empty vector, when rustflag is not specified?
Maybe you want something like this?

```suggestion
    let mut rustflags = matches.value_of(RUSTFLAGS).iter().collect::<Vec<_>>();
```

But i also don't really see the need to create a Vec for this instead of just appending to a String as it was before.

ogoffart (approved):
Looks good to me, with comments.

Feel free to rebase and merge.

generator/src/subcommands/build_crate.rs
jschwe: Thanks for catching this. You are right, as is, there is not really a reason to use a Vec. Reverted back to pushing to a String.

jschwe: > Feel free to rebase and merge.

Sorry about the unneccessary merge commit. Looks like I should have selected "Rebase and merge" in the drop down menu to avoid that. I assumed that github would automatically use fast forward merge if possible.

neachdainn:
Thanks for doing this!

PR #129, opened 2022-01-19
Correctly clean all the imported crates

jcaden:
When adding multiple crates, add all of them to the cargo-clean target


ogoffart:
Thank!

PR #130, opened 2022-01-22
Add --profile support for rust >= 1.57

jschwe:
Add the option to specify the cargo profile by optionally specifying
CARGO_PROFILE when importing a crate. Cargo profiles are specified on
the workspace level, so it makes sense to select this option with
corrosion_import_crate.
This can be useful for specifying e.g. a `release-with-dbg` or
`release-with-lto` profile etc, that the user can then conditionally
select.

We currently do appear to conditionally link some libraries, depending
on the CMake build mode. So if the user selects a profile that requires
libraries, that are only linked in a different CMake build mode, then
this may cause issues for them. But I would label that a configuration
issue, that can be improved separately if it turns out to be necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,16 @@ parameters for `corrosion_import_crate` called `ALL_FEATURES` and `NO_DEFAULT_FE
 boolean target properties - which override any specified values with `corrosion_import_crate` are called
 `CORROSION_ALL_FEATURES` and `CORROSION_NO_DEFAULT_FEATURES`.
 
-#### RUSTFLAGS
+### Selecting a custom cargo profile
+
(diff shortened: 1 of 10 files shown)


ogoffart:
Nice!

The patch looks good.

I believe it should be called simply `PROFILE`  instead of `CARGO_PROFILE`.  Or is `PROFILE` ambiguous?

jschwe:
>  Or is PROFILE ambiguous?

Someone could interpret it as in "performance profiling", but after thinking about it again `CARGO_PROFILE` doesn't really remove the ambiguity. I'll rename it tomorrow and adjust the documentation.

ogoffart:
Thanks

Issue #70, opened 2020-10-14
Potentially eliminate "corrosion-generator" when using CMake 3.19?


AndrewGaspar:
The main reason `corrosion-generator` exists is that it makes it easier to parse the `cargo metadata` output. CMake 3.19 adds a new [`string(JSON ...)`](https://cmake.org/cmake/help/latest/command/string.html#id2) command that makes parsing JSON much easier.

We'd definitely need to keep corrosion-generator for pre 3.19 targets, but potentially for 3.19 and later targets, it could be eliminated, making it much faster to use as a subdirectory.

PR #131, opened 2022-01-23
Support parsing cargo metadata by string(JSON ...) for cmake >= 3.19

crabtw:
New feature for #70

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -6,7 +6,19 @@ if (CMAKE_GENERATOR STREQUAL "Ninja Multi-Config")
         "https://github.com/corrosion-rs/corrosion/issues/50")
 endif()
 
+if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.19.0)
+    set(_CORROSION_CMAKE_GREATER_EQUAL_3_19 ON)
+else()
+    set(_CORROSION_CMAKE_GREATER_EQUAL_3_19 OFF)
+endif()
+
 option(CORROSION_VERBOSE_OUTPUT "Enables verbose output from Corrosion and Cargo" OFF)
+
+option(
+    CORROSION_PARSE_METADATA_BY_CMAKE
+    "Enable Corrosion to parse cargo metadata by cmake string(JSON ...) command"
+    ${_CORROSION_CMAKE_GREATER_EQUAL_3_19}
+)
 
 find_package(Rust REQUIRED)
 
@@ -23,6 +35,10 @@ get_property(
     CARGO_EXECUTABLE
     TARGET Rust::Cargo PROPERTY IMPORTED_LOCATION
 )
+
+if (_CORROSION_CMAKE_GREATER_EQUAL_3_19 AND CORROSION_PARSE_METADATA_BY_CMAKE)
+    include(CorrosionGenerator)
+endif()
 
 if (NOT TARGET Corrosion::Generator)
     set(_CORROSION_GENERATOR_EXE
@@ -185,7 +201,7 @@ function(_add_cargo_build)
     # can only add double quotes here. Any double quotes _in_ the rustflags must be escaped like `\\\"`.
     set(rustflags_genex "$<$<BOOL:${rustflags_target_property}>:--rustflags=\"${rustflags_target_property}\">")
 
-    if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.19.0)
+    if (_CORROSION_CMAKE_GREATER_EQUAL_3_19)
         set(build_env_variable_genex "$<GENEX_EVAL:$<TARGET_PROPERTY:${target_name},CORROSION_ENVIRONMENT_VARIABLES>>")
 
         set(features_target_property "$<GENEX_EVAL:$<TARGET_PROPERTY:${target_name},CORROSION_FEATURES>>")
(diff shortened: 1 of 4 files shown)


cmake/Corrosion.cmake
```diff
@@ -24,6 +24,10 @@ get_property(
     TARGET Rust::Cargo PROPERTY IMPORTED_LOCATION
 )
 
+if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.19.0)
```
jschwe, line 27: Since this is a rather large change, i would like to have CMake option to toggle this off in addition to the version check.

cmake/Corrosion.cmake
```diff
@@ -314,47 +322,68 @@ function(corrosion_import_crate)
         "${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_FILES_DIRECTORY}/corrosion/${toml_dir_name}.dir/cargo-build.cmake"
     )
 
-    if (CMAKE_VS_PLATFORM_NAME)
-        set (_CORROSION_CONFIGURATION_ROOT --configuration-root ${CMAKE_VS_PLATFORM_NAME})
-    endif()
+    if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.19.0)
```
jschwe, line 325: I guess you probably wanted the `_generator_gen_cmake()` in the true branch here? Or am I missing something

cmake/CorrosionGenerator.cmake
```diff
@@ -0,0 +1,519 @@
...
+                        --manifest-path ${manifest}
+                        --format-version 1
+        OUTPUT_VARIABLE json
+        COMMAND_ERROR_IS_FATAL ANY
+    )
+
+    set(${out} ${json} PARENT_SCOPE)
+endfunction(_cargo_metadata)
+
+function(_generator_print_root out manifest)
```
jschwe, line 25: The original rust command actually printed to stdout, but here there is no printing, so I think it would make sense to adjust the function name

cmake/CorrosionGenerator.cmake
```diff
@@ -0,0 +1,519 @@
...
+                ${CARGO_EXECUTABLE}
+                    metadata
+                        --manifest-path ${manifest}
+                        --format-version 1
+        OUTPUT_VARIABLE json
+        COMMAND_ERROR_IS_FATAL ANY
+    )
+
+    set(${out} ${json} PARENT_SCOPE)
+endfunction(_cargo_metadata)
```
jschwe, line 23: Since `CorrosionGenerator.cmake` requires version `3.19` anyway, I would remove the function name from endfunction. https://cmake.org/cmake/help/v3.19/command/endfunction.html

jschwe (commented):
First of all thanks for this PR! I have only skimmed the PR and will have a closer look and test it tomorrow, but I have left some comments already.

cmake/Corrosion.cmake
crabtw: done

cmake/Corrosion.cmake
crabtw: It's an embarrassing mistake, thanks.

cmake/CorrosionGenerator.cmake
crabtw: done

cmake/CorrosionGenerator.cmake
crabtw: done

jschwe:
I don't think generating CMake code with `file(APPEND ${out_file}` is actually necessary for the new CorrosionGenerator, since it is just included later anyway.
Unless I'm missing something, I would ask you to just write the code normally there, instead of writing to a file and then including it. This was necessary for the generator written in rust, but can be avoided here.

Also, I merged #130. so it would be great if you could also add the `PROFILE` option to the new CorrosionGenerator.

jschwe (commented):
Thanks for the changes! I've had time for a more thorough look now and I think it looks fine.
I left some more minor comments.

Also I would prefer if you would rename `CORROSION_PARSE_METADATA_BY_CMAKE` to `CORROSION_EXPERIMENTAL_PARSER` and add the option to the `README.md` under Usage->Options->Advanced explaining what the option does.

I tested the new generator with our project at work and it doesn't seem to break anything for us, so I'm fine with approving this MR. @ogoffart Do you want to do a second review?

crabtw:
@jschwe Done, thanks!

cmake/CorrosionGenerator.cmake
```diff
@@ -144,8 +144,8 @@ function(_generator_parse_target manifest package target)
         set(implib_name "${lib_name}.dll.lib")
     elseif(is_windows_gnu)
         set(implib_name "lib${lib_name}.dll.a")
-    else()
-        set(implib_name "lib${lib_name}.a")
+    else(is_windows)
```
jschwe, line 147: This one should be `elseif`

crabtw: Done, thanks!

housel:
This commit seems to break the `PROPERTY CORROSION_ENVIRONMENT_VARIABLES` feature for me on CMake 3.22

jschwe:
Hi @housel, thanks for the report! Could you open an issue describing in a bit more detail what problem you are facing? 
For me `CORROSION_ENVIRONMENT_VARIABLES` appear to work on CMake 3.22, and we also have the `envvar` test, which is passing.

PR #137, opened 2022-02-06
Allow Ninja Multi-config for Cmake >= 3.20

### LINKED ISSUE #50: Corrosion is incompatible with Ninja Multi-Config
When I try to build my project with Corrosion and the Ninja Multi-Config generator, I get a mesage like this:

```
$ cd build && cmake Ninja -G "Multi-Config" ..
[…]
-- Using Corrosion as a subdirectory
-- Configuring done
-- Generating done
-- Build files have been written to: /some/path/build

$ cmake --build . --
ninja: error: 'Debug/cargo/build/x86_64-unknown-linux-gnu/debug/libfoo.a', needed by 'src/bar/Debug/bar', missing and no known rule to make it
```

If I use `-G Ninja`, it works. I had this issue with single-conf vs. multi-conf. in my project as well. It usually boils doin to a few generator expressions at places where paths are constructed.

Thanks :-)

jschwe:
This commit is not extensively tested but at a first glance,
Ninja Multiconfig seems to work. I'll test this a bit more thoroughly at work tomorrow.

Closes #50

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -1,9 +1,8 @@
 cmake_minimum_required(VERSION 3.12)
 
-if (CMAKE_GENERATOR STREQUAL "Ninja Multi-Config")
-    message(FATAL_ERROR "Corrosion currently cannot be used with the \"Ninja Multi-Config\" "
-        "generator. Please use a different generator. Follow discussion on this issue at "
-        "https://github.com/corrosion-rs/corrosion/issues/50")
+if (CMAKE_GENERATOR STREQUAL "Ninja Multi-Config" AND CMAKE_VERSION VERSION_LESS 3.20.0)
+    message(FATAL_ERROR "Corrosion requires at least CMake 3.20 with the \"Ninja Multi-Config\" "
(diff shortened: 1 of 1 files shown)


cmake/Corrosion.cmake
```diff
@@ -240,9 +239,9 @@ function(_add_cargo_build)
     set(byproducts)
     foreach(byproduct_file ${ACB_BYPRODUCTS})
         list(APPEND build_byproducts "${cargo_build_dir}/${byproduct_file}")
```
ogoffart, line 241: What about this one?
There is also something similar line 154

jschwe: `cargo_build_dir` depends on both `build_dir` (which depends on $<CONFIG> for Multi-Config Generator) and on `build_type_dir`, which is either the cargo profile (if specified) or `debug` for debug builds or release for all other builds.

So the current behaviour is probably overly pessimistic and builds more seperately then it needs to.

jschwe:
I'll go ahead and merge this to master. There could of course still be some hidden issues, but since Ninja-Multiconfig is currently just disabled, I don't think this MR can really break anything.

PR #138, opened 2022-02-06
Set CC and CXX for cargo

### LINKED ISSUE #85: Map CMake compilers to cargo build environment variables
I tried to compile a project that failed because it used `c++` instead of the compiler specified in `CMAKE_CXX_COMPILER` (for example, `clang++`). I temporarly fixed this by making a link to `clang++` from `c++` in a location searched by $PATH before /usr/bin, but I think the better solution would be for corrosion to use the compiler specified by CMAKE_CXX_COMPILER.

jschwe:
Cargo build scripts may themselves build c-dependencies and may choose
the compiler by inspecting CC or CXX.
Thus we should pass the compiler that was set in CMake, or alternatively
that is set in the environment to cargo.
This allows the build script to select the same compiler.

Closes #85

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -252,6 +252,17 @@ function(_add_cargo_build)
         list(APPEND features_args --features ${feature})
     endforeach()
 
+    # Todo: Refactor this together with the "$compilers section above."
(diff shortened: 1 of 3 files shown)


ogoffart (approved):
Good idea

PR #139, opened 2022-02-06
Fix Integrator build script

### LINKED ISSUE #128: corrosion_link_libraries doesn't seem to handle multiple libraries
While `corrosion_link_libraries` seems to work with a single library, when there is more than one then:

- `${link_libs}` and `${search_dirs}` get expanded in the custom command as unquoted, space-separated values, which breaks the setting of the corresponding environment variables
- After adding quotes to fix the above, it still doesn't work because `corrosion_build::build_script();` is expecting a semicolon-separated list.

I'm using CMake 3.22.

jschwe:
We use `COMMAND_EXPAND_LISTS`, which also expands lists in generator
expressions. Therefore the values of the env variables will be space
seperated.

Closes #128

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -105,8 +105,13 @@ function(_add_cargo_build)
         set (build_dir .)
     endif()
 
-    set(link_libs "$<GENEX_EVAL:$<TARGET_PROPERTY:cargo-build_${target_name},CARGO_LINK_LIBRARIES>>")
-    set(search_dirs "$<GENEX_EVAL:$<TARGET_PROPERTY:cargo-build_${target_name},CARGO_LINK_DIRECTORIES>>")
+    # Join the list with `:` as a custom delimiter, since the `;` is expanded when passed
+    # to commands with COMMAND_EXPANDS_LISTS. File/directory paths may contain spaces, so we can't let
+    # the list be expanded.
+    set(_link_libs "$<GENEX_EVAL:$<TARGET_PROPERTY:cargo-build_${target_name},CARGO_LINK_LIBRARIES>>")
+    set(link_libs "$<JOIN:${_link_libs},:>")
+    set(_search_dirs "$<GENEX_EVAL:$<TARGET_PROPERTY:cargo-build_${target_name},CARGO_LINK_DIRECTORIES>>")
+    set(search_dirs "$<JOIN:${_search_dirs},:>")
 
     # For MSVC targets, don't mess with linker preferences.
     # TODO: We still should probably make sure that rustc is using the correct cl.exe to link programs.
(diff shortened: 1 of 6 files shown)


ogoffart:
Can there be a test?

Will there not be problems if the directories contains spaces? Or is that properly escaped. (Maybe that's another thing worth testing: spaces or special characters in the source and build paths)

jschwe:
> Will there not be problems if the directories contains spaces? Or is that properly escaped

I haven't tested, but I would imagine that to be problematic. I think there are probably also a couple of other places in our code that could be problematic for paths with spaces (but I'll look into that separately).
In this case, I guess that means we will have to add our own delimiter, because I don't think we can properly escape.

So that means I'll definitely also have to add a test.

jschwe:
I finally got back to this. I chose `:` as a delimiter, since it can=t be part of a filename, and added a test with multiple librariers including a filepath with spaces.

cmake/Corrosion.cmake
```diff
@@ -105,8 +105,13 @@ function(_add_cargo_build)
         set (build_dir .)
     endif()
 
-    set(link_libs "$<GENEX_EVAL:$<TARGET_PROPERTY:cargo-build_${target_name},CARGO_LINK_LIBRARIES>>")
-    set(search_dirs "$<GENEX_EVAL:$<TARGET_PROPERTY:cargo-build_${target_name},CARGO_LINK_DIRECTORIES>>")
+    # Join the list with `|` as a custom delimiter, since the `;` is expanded when passed
```
ogoffart, line 108: Did you mean `:` here?
Path can contain `:` on Linux iirc.

cmake/Corrosion.cmake
```diff
@@ -201,6 +206,7 @@ function(_add_cargo_build)
     # preserve any double quotes and keep it as one argument value. However single quotes don't work on windows, so we
     # can only add double quotes here. Any double quotes _in_ the rustflags must be escaped like `\\\"`.
     set(rustflags_genex "$<$<BOOL:${rustflags_target_property}>:--rustflags=\"${rustflags_target_property}\">")
+    #set(link_libs_rustflags_genex  "$<$<BOOL:${rustflags_target_property}>:--rustflags=\"${link_libs}\">")
```
ogoffart, line 209: Leftover?

test/cpp2rust/path with space/lib3.cpp
```diff
@@ -0,0 +1,10 @@
+// Check that libraries located at a path containing a space can also be linked.
```
ogoffart, line 1: I'd go with a even more complex path for the test. Include a few symbol, some emoji, an such.

ogoffart (approved):

cmake/Corrosion.cmake
jschwe: Yes, I changed to `:`, actually because PATH is `:` delimited. So using the same delimiter for our usecase seems reasonable. I adjusted the wording to make it clearer that file/directory paths cannot contain `:`.

cmake/Corrosion.cmake
jschwe: Yes :)

test/cpp2rust/path with space/lib3.cpp
jschwe: I added some chinese characters and emojis. I don't really want to add any of the ascii special characters, because then we would have to check what the common subset of supported special characters in filepath is across our supported operating systems and generators/compilers.

jschwe: Looks like ninja 1.11 (which is not even out yet) would be required to support non-ansi filepaths on windows. See https://github.com/ninja-build/ninja/pull/1915 for details.

Issue #133, opened 2022-02-03
Fix corrosion_set_linker_language


jschwe:
PR #126 brought up the topic that `corrosion_set_linker_language` currently does not do anything, since the used property `LINKER_LANGUAGE` is apparently special and can't be used in the way that it was.
Therefore the current behaviour of `corrosion_set_linker_language` is that nothing happens, and the choice of the linker is not influenced by using this function. E.g. adding `corrosion_set_linker_language(rust-exe C)` to `cpp2rust/CMakeLists.txt` does _not_ break the build, since still the C++ linker is invoked.

As far as I can see this topic is only relevant for x2rust style projects, where linking is invoked via cargo.

#126 proposed to fix this by renaming the used property. Due to other corrosion code, the proposed change has the effect that the language specified in `corrosion_set_linker_language` is appended to a list of `linker_languages` and `build_crate.rs` will choose the language from the list with the highest `CMAKE_<lang>_LINKER_PREFERENCE`. This is probably not what a user would expect from the name of the function (In the case of cpp2rust, again still the C++ linker is invoked).

We could of course also fix that, but I would instead propose to just deprecate `corrosion_set_linker_language` (which should not be a breaking change, since it effectively doesn't do anything anyway). Instead I think it would make more sense to add a function `corrosion_set_linker(<target> <linker>)`, which allows the user to directly specify the linker it wants to use. If you want to explicitely use the compiler/linker CMake chose for a certain language, you could still do `corrosion_set_linker(my_target ${CMAKE_C_COMPILER})` instead of explicitely specifiying a linker. 

The new behaviour would then be that if a linker was explictly set via `corrosion_set_linker`, then the `linker_languages` and `CMAKE_<lang>_LINKER_PREFERENCE` is just completely ignored and the linker specified by the user is passed to cargo via `CARGO_TARGET_<target_triple>_LINKER`

Any thoughts on this?

jschwe:
When adding a test for #139, I noticed that `corrosion_set_linker_language` apparently does have an effect. I'm not sure what exactly changed in comparison to my previous tests, so I'll need to investigate that further.

PR #140, opened 2022-02-06
Don't set linker language for corrosion generator

### LINKED ISSUE #88: Link with C compiler instead of C++ compiler
It's kind of stupid that we currently use the C++ compiler to link. This is done solely because we use the C++ compiler for the tests anyway. However, you're much more likely to have a C compiler than a C++ compiler, so it makes sense to use the C compiler instead.

jschwe:
The corrosion generator is a 100% rust executable, so I don't think there is any benefit in us telling rust which linker it should use.

Additionally corrosion_set_linker_language currently does not have any effect anyway (See #133).

Closes #88

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,8 +89,6 @@ endif()
 # Builds the generator executable
 corrosion_import_crate(MANIFEST_PATH generator/Cargo.toml)
 
-# Set the link language to CXX since it's already enabled
-corrosion_set_linker_language(corrosion-generator CXX)
 
 if (IS_ABSOLUTE ${CMAKE_INSTALL_LIBEXECDIR})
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -89,8 +89,6 @@ endif()
 # Builds the generator executable
 corrosion_import_crate(MANIFEST_PATH generator/Cargo.toml)
```
ogoffart, line 91:
```suggestion
```

ogoffart (commented):
If I'm not mistaken, corrosion_set_linker_language  is no longer used after this.
If that supposed to be used as a public function, it should be documented and somehow tested. But otherwise, should we simply remove it?

jschwe:
I think it is supposed to be a public function. I opened #133 to discuss removing it and offering something like `corrosion_set_linker()` instead. If you don't have any objections, I could just add that to the scope of this PR.

Issue #61, opened 2020-10-07
Document use of a bin crate in the build process


AndrewGaspar:
This is common for code generators - write a tool that lives in tree, then use that tool to generate code as part of the build. Will be used for certain configurations of bindgen support, as well.

RE: #60

Issue #60, opened 2020-10-07
Is it possible to use bin types?


kjeremy:
I'm looking into building some ROS nodes using a combination of corrosion and https://github.com/adnanademovic/rosrust

The examples show linking cpp and rust together but what I want to do is use corrosion to build my rust binary and then refer to it elsewhere in my cmake file (for instance in an install step).

AndrewGaspar:
Yep, that's definitely supported! You're going to want to do something like this:

```cmake
add_custom_command(
    OUTPUT my_output.txt
    COMMAND $<TARGET_PROPERTY:my-bin-crate,IMPORTED_LOCATION> -o my_output.txt
)
```

AndrewGaspar:
You might not need the generator expression if you're just using it for install, though. The generator expression is mostly to make sure that my-bin-crate is built before trying to build my_output.txt (it sets up the dependencies correctly). To customize install, you'll need to write an install script, which can then use your bin crate after you've built: https://cmake.org/cmake/help/latest/command/install.html?highlight=install#script

kjeremy:
I will have to read up on this (I am not a cmake expert). Typically for C++ projects using catkin this goes something like:

```cmake
cmake_minimum_required(VERSION 3.0.2)
project(publisher_node)

find_package(catkin REQUIRED)

catkin_package()

add_executable(${PROJECT_NAME}_node src/ros_test_node.cpp)
install(TARGETS ${PROJECT_NAME}_node
  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
)
```

I was hoping for a way to just refer to the crate output in `install(TARGETS`

AndrewGaspar:
Oh, I misunderstood. I thought you were trying to use the bin as part of install. You just want to install the bin.

Yeah, unfortunately CMake doesn't let you refer to imported targets in `install(TARGETS`. However, I added a custom function that behaves basically like `install(TARGETS` specifically for Corrosion targets. You can see an example usage in Corrosion itself: https://github.com/AndrewGaspar/corrosion/blob/master/CMakeLists.txt#L55

The function "signature" for corrosion_install is here: https://github.com/AndrewGaspar/corrosion/blob/master/cmake/Corrosion.cmake#L327

kjeremy:
[...]

PR #144, opened 2022-02-13
Test generated sources and calling find_package multiple times.

ratmice:
Here is a test which related to issue #60  and issue #61, and also tests my previous commit.

I'm not sure if it would be considered to fix those issues,  in that whether tests count as documentation,
one of those issues had mentioned `IMPORTED_LOCATION`, using `TARGET_FILE` instead seems stable across different `CONFIG`s.

It is worth noting that if you run this test with `Release` or `Debug` config rather than unspecified, it does output some warnings at cmake time. I include this warning below.

```
CMake Warning (dev) in gensource/srcgen/CMakeLists.txt:
  Policy CMP0111 is not set: An imported target missing its location property
  fails during generation.  Run "cmake --help-policy CMP0111" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  IMPORTED_LOCATION not set for imported target "srcgen".
This warning is for project developers.  Use -Wno-dev to suppress it.
```


ogoffart:
thanks

PR #149, opened 2022-02-25
Use corrosion-rs fork of install-cmake

jschwe:
To support CMake version >= 3.20 in CI, we should use the
updated fork. The download url format changed with cmake 3.20
so I updated the action to take that into account and return
the right url depending on the version.

I also saw there is the `get_cmake` action this action is partially
based on, but since that only supports one hardcoded version, it
would be hard to integrate into our CI workflow I think.


jschwe:
cherry-picking into #147 and closing this PR.

PR #154, opened 2022-03-25
Add NO_STD switch

### LINKED ISSUE #153: Support #[no_std] crates
Currently, when trying to add `no-std`-crates, several libraries are linked to the resulting binaries, causing several linking issues.
A way to disable this behavior would be great

arkkors:
Adds the NO_STD switch to Corrosion, allowing building
no_std crates in CMake without linking additional libraries.

Closes #153


ogoffart (approved):
Looks good!

I just wonder if the option should not be called just `NO_STD` to match the rust attribute.
Or is that a different thing?
Does it work for a no_std crate that uses the alloc crate?

arkkors:
1. Yeah, `NO_STD` is shorter :) Also, I don't really see the distinction (at least in practice).

2. I don't see why it shouldn't work with `alloc`, but I'm not entirely sure. Will try that out later.

arkkors:
1. The nostd test should now work under `1.44.1`, so the test suite should (hopefully) pass now.
2. Regarding the `alloc` crate question: The alloc crate itself works in this setup. You need to define a global allocator though and probably setup an alloc error handler (e.g via `#![feature(default_alloc_error_handler)]`). This needs a nightly toolchain AFAIK. But it works :)

jschwe:
Thanks for the PR. I see you also touched `gen_cmake/target.rs`. If possible, could you also add the same changes (basically add the include_platform_libs switch) to the experimental generator in at `cmake/CorrosionGenerator.cmake`? Should be this function: https://github.com/corrosion-rs/corrosion/blob/master/cmake/CorrosionGenerator.cmake#L25

arkkors:
Here you go. Not very pretty, but it works.

PR #156, opened 2022-03-27
Remove integrator buildscript

### LINKED ISSUE #54: Eliminate the build-script crate
It seems like we don't actually need the build-script crate - everything we need should be doable with CARGO_TARGET_<triple>_RUSTFLAGS.

jschwe:
Follow up to #139 

Remove the integrator buildscript and directly pass the linking arguments via `RUSTFLAGS`.

I'm keeping this as a separate PR (and commits) so that we can revert to the fixed build script in case removing the build script causes any problems.

Closes #54


ogoffart (approved):

jschwe:
@ogoffart I probably should add that this is a breaking change, since I just removed the build script crate. I would just schedule this change for the next major release 0.2, since it is pretty easy to fix on the user side and overall an improvement.

If you want to, I guess we could still keep the build script crate for a while. I don't think it really makes sense to add a deprecation notice though since the output of build scripts is usually not visible.

PR #157, opened 2022-03-28
Do not include libraries with NO_STD

arkkors:
There was a small mistake with #154: when setting NO_STD, the `--no-default-libraries` switch was not passed to the cmake generator This MR fixes the issue.


ogoffart:
thanks!

PR #158, opened 2022-04-01
Support Features, EnvVars and Hostbuild pre Cmake 3.19

jschwe:
CMake before v3.19 required the properties on INTERFACE libraries
to be prefixed with INTERFACE. Since previously we already
exposed some prefixed property names in our public api, we now
conditionally use a different property name to support older
CMake versions without breaking existing code and hide this behind a function.

The Property names are removed from the public API and
directly accessing them is considered deprecated. Instead
the newly introduced functions should be used.

Todo: ~~Change the Readme to reflect the API changes and document support of older versions~~


tronical:
First of all: I think it's terrific how you guys are maintaining and improving corrosion.

@ogoffart asked me to chime in here. I think @hunger 's feedback would also be valuable here.

I think in CMake itself most of the APIs nowadays are property based if possible, i.e. CMake adds new features often as properties, with different scope (target, global, per-directory, etc.). Adding new functions for new build features seems the exception to me.

In the light of that I think a property based API makes also sense for corrosion to maintain, as far as that's possible. Clearly the crate import functionality requires a function :-).

Regarding that, I recall we faced a similar situation when designing in the Qt 6 CMake APIs, where functions that ended up creating new CMake targets ended up with a function name that resembles the CMake pattern, so for example `qt_add_executable`, `qt_add_qml_module`, etc.

Regarding the existing patch, I'm also curious about one detail: `corrosion_set_features(TARGET <target_name>` takes a `TARGET` while the other functions don't. Is there a specific reason for this? Do you intend to extend `corrosion_set_features` to accept something other than a target as the first parameter in the future

jschwe:
Hi,

Thanks for your feedback!

> I think in CMake itself most of the APIs nowadays are property based if possible, i.e. CMake adds new features often as properties, with different scope (target, global, per-directory, etc.). Adding new functions for new build features seems the exception to me.

I'm not really  a CMake expert, but I think the main thing to avoid in "modern" CMake is to using functions which affect the global scope. I think adding functions which only affect the target scope should still be fine.

In the case of this specific patch, target properties are actually still what is used in the background - I just want to hide the name. This is due to us previously exposing the property name as part of our public API, but this also prevents compatability with previous versions of CMake (custom properties on INTERFACE targets needed to be prefixed with `INTERFACE` before CMake 3.19. Now this leaves the following options for supporting older cmake versions:
- Rename the property (breaking change).
- Make the property name depend on the CMake version and document both names
- Make the property name depend on the CMake version and hide the name by hiding it behind a function.

I personally like the last option best, because it doesn't break existing code and at the same time doesn't overcomplicate the documentation. I'm not really a fan of the second option, but I would be open to renaming the property for everyone.

> Regarding the existing patch, I'm also curious about one detail: corrosion_set_features(TARGET <target_name> takes a TARGET while the other functions don't. Is there a specific reason for this? Do you intend to extend corrosion_set_features to accept something other than a target as the first parameter in the future

No, there is actually no reason for this, except that I needed to use cmake_parse_arguments anyway, so I arbitrarily decided to just make target_name a named parameter. I guess it is inconsistent, and I don't see the target_name becoming optional in the future, so I think it makes sense to change it into a positional parameter.

tronical:
Thanks for the explanations :)

> I'm not really a CMake expert, but I think the main thing to avoid in "modern" CMake is to using functions which affect the global scope. I think adding functions which only affect the target scope should still be fine.

I agree, that is certain an aspect of "modern" CMake.

> I personally like the last option best, because it doesn't break existing code and at the same time doesn't overcomplicate the documentation. I'm not really a fan of the second option, but I would be open to renaming the property for everyone.

You make a compelling point (that I evidently missed before). I agree that functions are good way to support older CMake versions. Since old code shouldn't break, I think it makes sense to still document the names that are used in newer CMake versions.

I guess the other option is to bump the minimum CMake version.

hunger:
> I'm not really a CMake expert,

I have yet to meet somebody bold enough to make that claim about themselves ;-)

> [...] but I think the main thing to avoid in "modern" CMake is to using functions which affect the global scope. I think adding functions which only affect the target scope should still be fine.

Yeap, this is a nice enough way to get compatibility going forward, by abstractring over the abstraction layer.

> * Rename the property (breaking change).

This is what I would personally go for, in total ignorance of your audience:-)

> * Make the property name depend on the CMake version and document both names

This is what I would avoid by all means necessary.

> * Make the property name depend on the CMake version and hide the name by hiding it behind a function.

This will work and get you a nice abstraction layer on top of the abstraction layer called CMake.

You forgot the option to just bump CMake version and use the new names unconditionally, which would be my preferred option. In my experience you need to follow upstream CMake pretty closely when you want to integrate with it. That way you can just fix issues upstream instead of working around problems downstream. The CMake people are very careful about adding new features and try hard not to break things for their users, so this is not as painful as it initially sounds:-)

But I am mostly ignorant about your audience, so I do not know how applicable my opinion is to the topic at hand.

jschwe:
> Since old code shouldn't break, I think it makes sense to still document the names that are used in newer CMake versions.

I'm not sure I understand your point here. Old code still works (since the property name did not change for CMake >=3.19), regardless of the property name still being documented or not. 

> I guess the other option is to bump the minimum CMake version.

> You forgot the option to just bump CMake version and use the new names unconditionally, which would be my preferred option.

Well, that is basically the status quo. This PR is just about "backporting" existing functionality to previous CMake versions.
The CMake versions shipped with Ubuntu are 3.18 for 21.10 and 3.16 for 20.04 LTS. I think the most recent Debian even only ships 3.13. 
I would like to reasonably support at least the cmake version shipped with Ubuntu 20.04 LTS release for now. Setting environment variables and selecting cargo features seem pretty essential to me. Until now this requires CMake 3.19, simply because target property names on INTERFACE libraries must be prefixed with `INTERFACE`.

> In my experience you need to follow upstream CMake pretty closely when you want to integrate with it. That way you can just fix issues upstream instead of working around problems downstream

I generally agree with this, but in this case I feel that conditionally renaming a property to be prefixed with `INTERFACE` is a pretty small inconvenience to pay.

> But I am mostly ignorant about your audience, so I do not know how applicable my opinion is to the topic at hand.

I actually also have no idea who uses corrosion and what cmake versions they use.

PR #165, opened 2022-04-12
Don't generate .cargo/config

jschwe:
It's better to pass the arguments directly on the commandline.

This is a first step towards allowing the working directory for cargo
build to be set by the user, which would allow the user to use a
.cargo/config.toml themselves


ogoffart (approved):
looks sensible

PR #167, opened 2022-04-12
Fix macos-11

jschwe:
For Rust targets, Rust will invoke the linker and add (among others) the
 -lSystem flag. This works fine for the default linker, but if we use a
  custom linker (e.g. g++, or even /usr/bin/ld), then we must provide
  that path to the System library.

Adding -L<path> to RUSTFLAGS works fine for regular libraries, but fails
 for build-scripts (See https://github.com/rust-lang/cargo/issues/4423 )
because there is no way to set RUSTFLAGS for build scripts when
cross-compiling. This is mostly an issue because we always specify the
target, even when building for host, but that is already sufficient for
the RUSTFLAGS to not apply to the build-script-build.

Target specific RUSTFLAGS (i.e CARGO_TARGET_<triple>_RUSTFLAGS also
don't seem to affect buildscripts.

Setting LIBRARY_PATH=<path> applies to normal targets and build-scripts
too, and seems to fix the issue.

My previous fix probably only fixed the case where a rust library gets
linked into a foreign library or executable.


jayvdb (approved):
This fixed installation for me.  yay.

Issue #152, opened 2022-03-22
`string.h` file not found for macOS


niyue:
# Setup
I use corrosion to build a C++ project with some Rust dependency, and the Rust dependency has a transitive dependency to this crate https://github.com/MaterializeInc/rust-krb5-src

# my environment
Operating system: macOS 12.3

```
❯ cc --version
Apple clang version 13.1.6 (clang-1316.0.21.2)
Target: x86_64-apple-darwin21.4.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

# Issue
The above setup used to work for corrrosion f679545a63a8b214a415e086f910126ab66714fa, and I recently upgraded to a new version corrosion 0.1 release, and it starts to report the following error, and I tried a new commit (904fefbedff0a781ea6db861100f584326adb19c), which reports the same error as well. But if I rolled back to use f679545a63a8b214a415e086f910126ab66714fa, the error won't be reported. 

Could some one shed some light on it? Any help is appreciated. Thanks.

```
Compiling strip-ansi-escapes v0.1.1
error: failed to run custom build command for `krb5-src v0.3.2+1.19.2`
Caused by:
process didn't exit successfully: `/<redated>/x64-osx-dbg/cargo/build/debug/build/krb5-src-9aacd81c281a6b62/build-script-build` (exit status: 101)
--- stdout
checking for gcc... /Library/Developer/CommandLineTools/usr/bin/cc
checking whether the C compiler works... no
 --- stderr
configure: error: in `/<redated>/x64-osx-dbg/cargo/build/x86_64-apple-darwin/debug/build/krb5-src-6053a8bec92c223c/out/build':
configure: error: C compiler cannot create executables
See `config.log' for more details
thread 'main' panicked at 'configure failed: Custom { kind: Other, error: "command [\"./krb5/src/configure\", \"--prefix=/<redated>/x64-osx-dbg/cargo/build/x86_64-apple-darwin/debug/build/krb5-src-6053a8bec92c223c/out/install\", \"--enable-static\", \"--disable-shared\", \"--disable-nls\", \"CPPFLAGS=\", \"CFLAGS= -fPIC\"] exited with code 77" }', /Users/ss/.cargo/registry/src/github.com-1ecc6299db9ec823/krb5-src-0.3.2+1.19.2/build.rs:86:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
[...]

PR #170, opened 2022-04-15
Fix compilation of projects using cc-rs on macOS

tronical:
After commit 0b0b182f8d2e12a1558d12dc5aea15bc815c703b, cc-rs now invokes
the compiler found by CMake, for example

    /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc

instead of

    /usr/bin/cc

The latter uses the default SDK sysroot root as configured with xcode.
The latter requires either specifying --sysroot /path/to/sysroot (which
is what CMake does) or it falls back to using the SDKROOT environment
variable. Without a sysroot the compiler can't find standard includes.

This patch sets SDKROOT as environment variable to be set during cargo
invocation, to the same value that CMake passes to the compiler. That
way cc-rs won't need any additional flags.

cc #152

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -278,6 +278,13 @@ function(_add_cargo_build)
     if(DEFINED ENV{CXX})
         list(APPEND corrosion_cc_rs_flags "HOST_CXX=$ENV{CXX}")
     endif()
+    # Since we instruct cc-rs to use the compiler found by CMake, it is likely one that requires also
+    # specifying the target sysroot to use. CMake's generator makes sure to pass --sysroot with
+    # CMAKE_OSX_SYSROOT. Fortunately the compilers Apple ships also respect the SDKROOT environment
+    # variable, which we can set for use when cc-rs invokes the compiler.
+    if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Darwin" AND CMAKE_OSX_SYSROOT)
+        list(APPEND corrosion_cc_rs_flags "SDKROOT=${CMAKE_OSX_SYSROOT}")
+    endif()
 
     corrosion_add_target_rustflags("${target_name}" "$<$<BOOL:${corrosion_link_args}>:-Clink-args=${corrosion_link_args}>")
 


jschwe:
This makes sense to me. 
However I think this fix may not work for the scenario where cmake is configured for cross-compiling and the "Host build" option is set for a crate. Is it possible to adjust the fix to consider this scenario?

tronical:
> However I think this fix may not work for the scenario where cmake is configured for cross-compiling and the "Host build" option is set for a crate.

Can you elaborate how setting `SDKROOT` may break cross-compiling?

This environment variable is only used by the Apple clang compilers, so if cross-compiling to a different macOS version and architecture, then CMAKE_OSX_SYSROOT should be set correctly by the user of CMake, I think (or rather, cmake sets it correctly based on the deployment target).

jschwe:
Cross compiling would work fine. The issue appears when you are cross-compiling, but mark one crate as "Host build", i.e. don't cross-compile this specific crate. In that case the CMAKE_HOST variables would probably have to be used to get the correct information. 

On the other hand I'm not sure why we have the Host build option at all. I think users could just use build scripts to achieve build time code generation, but maybe I'm overlooking something.

tronical:
Ahh, yes, now I get it. You're right! I changed the patch.

Regarding the host build feature: We have this use-case that a CMake project consists of a library and a separate tool (a little compiler). Traditionally with CMake, cross-compiling in such a scenario requires building the entire project for the host first - in order to get a tool binary that can run - and then building the project another time, into a different prefix, in order to get the library for the target. The second build then may need some extra trickery to import the tool from the first build, to make it appear as seamless in usage as if it were a host == target build.

With cargo this is fortunately not necessary, we can invoke `cargo build -p libproj --target=foo` and `cargo build -p tool`  or - for pure Rust projects - use the tool via proc macros and it'll be correctly build for the host arch.

The corrosion host build feature brings that convenience to a CMake "wrapper" and eliminates the need for two cmake builds.

That said, in Slint - where we use this - we could also do without and go two a two-stage build system. I think eventually we may need that anyway for other reasons. So this is merely my explanation of the past, I have no objections if you want to remove this feature again in a future Corrosion release :-)

jschwe:
> Ahh, yes, now I get it. You're right! I changed the patch.

Thanks. This new fix will not be sufficient for the case where you [cross-compile from e.g. Linux to Mac](https://wapl.es/rust/2019/02/17/rust-cross-compile-linux-to-macos.html), but I guess that is pretty exotic anyway, so I'm fine with merging this PR as is.

> Regarding the host build feature: [...]

Thanks for elaborating on that! We actually have a similar use-case at work (we also cross-compile), where we have a tool that generates code. But since it's written in 100% Rust, we just do `add_custom_target(tool COMMAND ${CARGO_COMMAND} build --target-dir [...])` without involving Corrosion at all. 

For now, I don't mind leaving the host build feature in though.

PR #175, opened 2022-04-27
Respect OUTPUT_NAME when linking targets that have it defined

### LINKED ISSUE #172: corrosion_link_libraries doesn't respect the OUTPUT_NAME property
I was trying to link a rust crate to c++ libraries that were defined in my project, and came across the todo in corrosion_link_libraries

```cmake
        # TODO: The output name of the library can be overridden - find a way to support that.
        corrosion_add_target_rustflags(${target_name} "-l${library}")
```

This is exactly my problem, my libraries are defined as `add_library(LibFoo SHARED ${some_sources})` and then the actual filename on the filesystem is set like `set_target_properties(LibFoo PROPERTIES OUTPUT_NAME foo)`

I found that swapping out the actual link line with
```cmake
        if (TARGET ${library})
            corrosion_add_target_rustflags(${target_name} "-l$<TARGET_PROPERTY:${library},OUTPUT_NAME>")
        else()
            corrosion_add_target_rustflags(${target_name} "-l${library}")
        endif()
```

seems to do the trick, even if it's a bit ugly.

Would a patch like this be accepted as-is? What kind of testing would you want to confirm it works in more situations than the current code?

ADKaster:
This causes the OUTPUT_NAME target_property to be used when passing the
-l flag to corrosion_link_libraries. However, if any users have
corrosion_link_libraries(MyRustExe LibA) in their CMakeLists before the
target LibA is defined, it will now error out because the target is not
defined yet.

Fixes #172 

I'm not sure if this technically breaking change is ok. It definitely seems non-idiomatic to me to link against a library before it's been declared as a target, especially if you're going to set OUTPUT_NAME on it or other properties... but I can go back and wrap the get_property in an if(TARGET ${library}) if it's preferred.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -597,10 +597,9 @@ function(corrosion_link_libraries target_name)
             PROPERTY CARGO_DEPS_LINKER_LANGUAGES
             $<TARGET_PROPERTY:${library},LINKER_LANGUAGE>
         )
-        corrosion_add_target_rustflags(${target_name} "-L$<TARGET_LINKER_FILE_DIR:${library}>")
 
-        # TODO: The output name of the library can be overridden - find a way to support that.
-        corrosion_add_target_rustflags(${target_name} "-l${library}")
+        corrosion_add_target_rustflags(${target_name} "-L$<TARGET_LINKER_FILE_DIR:${library}>")
+        corrosion_add_target_rustflags(${target_name} "-l$<TARGET_LINKER_FILE_BASE_NAME:${library}>")
     endforeach()
 endfunction(corrosion_link_libraries)
 
--- a/test/cpp2rust/CMakeLists.txt
+++ b/test/cpp2rust/CMakeLists.txt
@@ -1,5 +1,4 @@
 corrosion_import_crate(MANIFEST_PATH rust/Cargo.toml)
-corrosion_link_libraries(rust-exe cpp-lib cpp-lib2 cpp-lib3)
 
 add_library(cpp-lib lib.cpp)
 target_compile_features(cpp-lib PRIVATE cxx_std_14)
@@ -15,6 +14,7 @@ set_target_properties(
         cpp-lib2
         PROPERTIES
         POSITION_INDEPENDENT_CODE ON
+        OUTPUT_NAME cpp-lib-renamed
 )
 
 add_library(cpp-lib3 "path with space/lib3.cpp" )
@@ -24,3 +24,5 @@ set_target_properties(
         PROPERTIES
         POSITION_INDEPENDENT_CODE ON
 )
+
+corrosion_link_libraries(rust-exe cpp-lib cpp-lib2 cpp-lib3)


cmake/Corrosion.cmake
```diff
@@ -587,10 +587,18 @@ function(corrosion_link_libraries target_name)
...
+            LIBRARY_OUTPUT_NAME
+            TARGET ${library}
+            PROPERTY OUTPUT_NAME
+        )
+        if (NOT LIBRARY_OUTPUT_NAME)
+            set(LIBRARY_OUTPUT_NAME ${library})
+        endif()
+
+        corrosion_add_target_rustflags(${target_name} "-L$<TARGET_LINKER_FILE_DIR:${library}>")
+        corrosion_add_target_rustflags(${target_name} "-l${LIBRARY_OUTPUT_NAME}")
```
jschwe, line 601: Couldn't we use [`$<TARGET_LINKER_FILE:tgt>`](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:TARGET_LINKER_FILE) here?

Looking at [LIBRARY_OUTPUT_NAME](https://cmake.org/cmake/help/latest/prop_tgt/LIBRARY_OUTPUT_NAME.html), it could be not set, but `OUTPUT_NAME` could be set. Using the `TARGET_LINKER_FILE` genex should probably automatically select the right version.

ADKaster: hm. When I use that instead, it fails to link the test.

````
"-L" "/home/andrew/corrosion/build/cargo/build/release/deps" "-L" "/home/andrew/corrosion/build/test/cpp2rust" "-L" "/home/andrew/corrosion/build/test/cpp2rust" "-L" "/home/andrew/corrosion/build/test/cpp2rust" "-L" "/home/andrew/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-l/home/andrew/corrosion/build/test/cpp2rust/libcpp-lib.a" "-l/home/andrew/corrosion/build/test/cpp2rust/libcpp-lib-renamed.a" "-l/home/andrew/corrosion/build/test/cpp2rust/libcpp-lib3.a"  <SNIP> "-Wl,--gc-sections" "-pie" "-Wl,-zrelro,-znow" "-Wl,-O1"
  = note: /usr/bin/ld: cannot find -l/home/andrew/corrosion/build/test/cpp2rust/libcpp-lib.a
          /usr/bin/ld: cannot find -l/home/andrew/corrosion/build/test/cpp2rust/libcpp-lib-renamed.a
          /usr/bin/ld: cannot find -l/home/andrew/corrosion/build/test/cpp2rust/libcpp-lib3.a
          collect2: error: ld returned 1 exit status
````
but you're right, my solution does ignore LIBRARY_OUTPUT_NAME, ARCHIVE_OUTPUT_NAME and the `` _<CONFIG>`` variants, which needs fixing

Comment (ADKaster) [reply cmake/Corrosion.cmake:601]: If CMake 3.15 is an ok base reqiurement, the TARGET_FILE_BASE_NAME genex could be used.

https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:TARGET_FILE_BASE_NAME

Comment (jschwe) [reply cmake/Corrosion.cmake:601]: I think you need to add a `:` like `-l:/path/x.a`, see the [documenation](https://linux.die.net/man/1/ld) for `--library`. Not 100% sure this works though.

> If CMake 3.15 is an ok base reqiurement

In that case this either needs to be guarded by a version check, or wait until after Corrosion v0.2 is released (which I plan to do soon, I just have to update the readme to reflect the recent changes).

Comment (ADKaster) [reply cmake/Corrosion.cmake:601]: I've poked around with this for about an hour now and can't figure out a proper way to do it that only requires the current minimum cmake of 3.12. putting `:` in the `-l` command screws up the INTERFACE_CORROSION_RUSTFLAGS generator expression and causes a funky compile error: 

e.g.
`
-- IFACE flags are: $<$<BOOL:$<$<NOT:$<BOOL:$<TARGET_PROPERTY:rust-exe,CORROSION_USE_HOST_BUILD>>>:>>:-Clink-args=$<$<NOT:$<BOOL:$<TARGET_PROPERTY:rust-exe,CORROSION_USE_HOST_BUILD>>>:>>;-Cdefault-linker-libraries=yes;-l:/home/andrew/corrosion/build/test/cpp2rust/libcpp-lib.a;-L$<TARGET_LINKER_FILE_DIR:cpp-lib2>;-l$<TARGET_LINKER_FILE_BASE_NAME:cpp-lib2>;-L$<TARGET_LINKER_FILE_DIR:cpp-lib3>;-l$<TARGET_LINKER_FILE_BASE_NAME:cpp-lib3>
`

```
error: renaming of the library `` was specified, however this crate contains no `#[link(...)]` attributes referencing this library
```

Similarly, trying to use the TARGET_LINKER_FILE genex but strip out the full path to just be the `get_filename_component(blah NAME_WE)` or such doesn't work, because the genex is expanded much later than configure time. 

So waiting for 0.2 to drop seems reasonable to me :) 

The symmetry of 
```
       corrosion_add_target_rustflags(${target_name} "-L$<TARGET_LINKER_FILE_DIR:${library}>")
        corrosion_add_target_rustflags(${target_name} "-l$<TARGET_LINKER_FILE_BASE_NAME:${library}>")
```
is pretty appealing at least.

jschwe: Thanks for taking the time to investigate! 

> So waiting for 0.2 to drop seems reasonable to me :)

Ok, great.

ADKaster: I've pushed an update to use the preferred syntax, but given your announcement it will have to wait until v0.3 :)

jschwe:
I'll bump the minimum version in the CI and in the documentation to 3.15 later.

cmake/Corrosion.cmake
jschwe: @ADKaster I've updated the minimum supported CMake version to 3.15. Could you rebase your changes?

ADKaster: Sure thing, rebased on master :)

jschwe (approved):

jschwe:
Thanks for this PR.

PR #177, opened 2022-05-01
Update Readme and "stabilize" option for disabling native tooling

jschwe:
- Update Readme to reflect recent changes and fix some formatting.
- Rename the `CORROSION_EXPERIMENT_PARSER`Option: For the upcoming v0.2 release, the cmake tooling repacing the native tooling should no longer be considered experimental. To reflect the legacy status of the native tooling better, rename and invert the option semantics. In order to not break anything for people using corrosion on the master branch, also keep support for the original option, but add a deprecation notice, so that we can safely remove it in the next version.


README.md
```diff
@@ -89,14 +91,15 @@ environment variable. This is likely to already be the case by default on a Unix
...
-Once Corrosion is installed and you've ensured the package is avilable in your `PATH`, you
+Once Corrosion is installed and you've ensured the package is available in your `PATH`, you
 can use it from your own project like any other package from your CMakeLists.txt:
 ```cmake
 find_package(Corrosion REQUIRED)
 ```
 
 ### FetchContent
-If installation is difficult or not feasible in your environment, you can use the
+If you are using a recent CMake version (>= 3.19) or installation is difficult or not feasible in
```
ogoffart, line 101:
```suggestion
If you are using a CMake >= 3.19 or installation is difficult or not feasible in
```

"recent" is fairly relative, just say what version.

But i'm not sure i understand why there is a `or` here.  Should it be a `and`? (why do we recommand this with CMake .19 but not with earlier version.)

README.md
```diff
@@ -107,7 +110,7 @@ include(FetchContent)
 FetchContent_Declare(
     Corrosion
     GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git
-    GIT_TAG origin/master # Optionally specify a version tag or branch here
+    GIT_TAG origin/master # Optionally specify a commit hash, version tag or branch here
```
ogoffart, line 113: should we actually use the latest stable tag here in that doc so people that copy paste from there get a stable setup?

ogoffart (approved):
These changes looks nice. thanks.

README.md
jschwe: Makes sense to me. I set it to v0.1.0 for now.

README.md
jschwe: > "recent" is fairly relative, just say what version.

Done

Regarding the `or`:  For Cmake >= 3.19 we don't need the native tooling, so the configure is equally fast regardless of corrosion being installed or corrosion being used as a subdirectory. For older CMake versions we need the native generator tool, which increases the configure time significantly if corrosion is not installed.

PR #181, opened 2022-05-06
Fix usage of corrosion via add_subdirectory by promoting variables to CACHE

lxbrz:
Some of the variables that corrosion provides for its users like
`Rust_VERSION<_MAJOR|_MINOR|_PATCH>` are only available when corrosion is installed and imported
via `find_package(Corrosion)`. This is because these variables are set as normal variables with
CMakes set() command, which limits their scope. When corrosion is imported via `add_subdirectory`,
these variables were only visible to corrosion, but not to the caller of `add_subdirectory`, which
is fixed by this commit.


jschwe:
Thanks for fixing this. Looks good to me.

jschwe (approved):

Issue #183, opened 2022-05-18
Is option std_cxx_17 required ?


stephan57160:
I failed to rebuild `corrosion` on CentOS-7.x-based system.
One of the reasons is the rather old compiler versions on this system.
Please, don't laugh, but I use this old stuff for many years and other projects.

Anyway, in this context `corrosion` build process fails with
```
CMake Error at test/cpp2rust/CMakeLists.txt:5 (target_compile_features):
  target_compile_features The compiler feature "cxx_std_17" is not known to
  CXX compiler

  "GNU"

  version 4.8.5.
```
...

In `test/cpp2rust/CMakeLists.txt`, I changed `std_cxx_17` to `std_cxx_14`.
With the help of some minor other things, I managed to have compilation till the end.

Before I raise a PR, I'd juste like to know if there is any reason for this CMAKE option ?
Is it a requirement from RUST, or from anything else ?
Would this break anything, if removed ?

ogoffart:
This is just needed for this one test. It uses string_view which is C++17.

You shouldn't need C++17 to use corrosion though?
How are you using Corrosion? Are you explicitly turning on the tests?

stephan57160:
Well ... I plan to use it, yes, but still in evaluation phase, with some other things.

I came to this, with the "build" process given in the README, nothing more, nothing less.

By the way, the minor other things I mentioned, are related to <string_view> in the 3 test cases :-)
Had to drop its use in `lib.cpp`, `lib2.cpp` and `lib3.cpp` ...

If really not required for `corrosion` to work, I may send a PR.
Just tell me, else, I keep the patch for myself, at least, for now.

ogoffart:
I think maybe the problem is that the tests are build by default if corrosion is built from the root directory.  Maybe `CORROSION_BUILD_TESTS` should always default to false, and we should document to enable the tests for developer wanting to hack on corrosion.

In practice though, if you simply want to use corrosion, you probably will want to use `FetchContent` or bring corrosion in a sub-directory.   In that case, the tests are not build by default.


Anyway, regarding your direct question, I think it'd be alright to port the tests away from C++17.

jschwe:
Another way would be to use CMake to test if 
a) There is a C++ compiler on the system, and
b) If the C++ compiler is recent enough
and skip the test in question if that is not the case.
[...]

PR #184, opened 2022-05-19
Corrosion tests should not test C++ features not used by Corrosion it…

stephan57160:
…self.

This PR follows discussion for #183.

* Problem addressed with this PR:
- CMakeLists.txt uses std_cxx_17 to compile it's test cases.
- Tests cases need std_cxx_17, because of their use of <string_view>.
Because of both, Corrosion fails to compile on 'old' systems (I did not say vintage :-)).

* Investigations:
As Corrosion does not use <string_view>, it's not needed to use this in its test cases.
If <string_view> is not used anywhere, options std_cxx_17 are no more required.

* Correction proposal:
Use of std_cxx_14, instead of std_cxx_17.
Remove the use of <string_view> in Corrosion test cases.
Altogether, this allows older compilers to build Corrosion.

* Remark:
This PR keeps some std_cxx_14 (instead of _17) to keep the need of C++ Standard Library installed on the compilation host.
I did not test, but I suppose std_cxx_11 would work too.

### DIFF
--- a/test/cpp2rust/lib.cpp
+++ b/test/cpp2rust/lib.cpp
@@ -1,5 +1,5 @@
 #include <iostream>
 
 extern "C" void cpp_function(char const *name) {
-    std::cout << "Hello, " << "my name is lib.cpp" << "! I'm C++!\n";
+    std::cout << "Hello, " << name << "! I'm C++!\n";
 }
--- a/test/cpp2rust/lib.cpp
+++ b/test/cpp2rust/lib.cpp
@@ -1,7 +1,5 @@
 #include <iostream>
-#include <string_view>
 
 extern "C" void cpp_function(char const *name) {
-    std::string_view const name_sv = name;
-    std::cout << "Hello, " << name_sv << "! I'm C++!\n";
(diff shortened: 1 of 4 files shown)


test/cpp2rust/lib.cpp
```diff
@@ -1,7 +1,5 @@
 #include <iostream>
-#include <string_view>
 
 extern "C" void cpp_function(char const *name) {
-    std::string_view const name_sv = name;
-    std::cout << "Hello, " << name_sv << "! I'm C++!\n";
+    std::cout << "Hello, " << "my name is lib.cpp" << "! I'm C++!\n";
```
ogoffart, line 4:
```suggestion
    std::cout << "Hello, " << name << "! I'm C++!\n";
```

ogoffart (changes requested):
Looks good.  But we should probably still print the name coming from rust, even if not wrapped in a string_view.

test/cpp2rust/lib.cpp
stephan57160: Oops :-)
I'll change that. Sorry.

What do you prefer ? 
A 2nd `commit` or a `squash` ?

Issue #168, opened 2022-04-13
Feedback wanted: Increase the Minimum supported Cmake version


jschwe:
The Minimum CMake version will likely be increased for `v0.3`. At the very least I want to drop support for CMake 3.12, but I'm also considering requiring CMake 3.16 or even 3.18. If you are using a CMake version that would be no longer supported by corrosion, please comment here so that I can gauge the number of affected users.

This issue will be mentioned in the Release notes for `v0.2`, so everyone who reads release notes should have the opportunity to voice themselves.

heinrich5991:
Debian 10 (without backports) includes CMake 3.13.4 (https://repologoy.org/project/cmake/versions). Debian 10 is still supported until 2024 (https://en.wikipedia.org/w/index.php?title=Debian_version_history&oldid=1089641928#Release_table).

Would you accept patches for backward compatibility in case I start using this? My project tries to support Debian.

heinrich5991:
No, I probably don't need new features.

jwoolston:
We use corrosion in a mixed x86/Arm environment. The Arm builds are with Buildroot versions which currently limit us to CMake 3.16 which is equivalent to the "standard" provided by Ubuntu 20.04 (our official development environment).

PR #185, opened 2022-05-22
Increase Minimum required Cmake version to 3.15

jschwe:
- CMake 3.15 brings a number of improvements regarding logging.
- Required to properly support OUTPUT_DIRECTORY and co.
- Remove 3.18 from the test matrix, since testing 3.15 should be ~~equivalent~~ close enough,
as the special cases for 3.12 are no longer there.

Issue https://github.com/corrosion-rs/corrosion/issues/168 is still relevant though, since if no one objects, it may make sense to raise the minimum supported CMake version even higher before the next release.


jschwe:
I'm not sure if we should also bump the minimum required version for `FindRust`, since I think some people actually use `FindRust`, but not corrosion.

ogoffart (approved):
> I'm not sure if we should also bump the minimum required version for FindRust, since I think some people actually use FindRust, but not corrosion.

Then let's not bump it unless there is a reason to do so.

PR #193, opened 2022-07-28
Use macos-12 in github actions

jschwe:
macos-10.15 is deprecated and will be removed at the end of next month
according to this announcement: https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/

Use the newest available version macos-12 in github actions instead.

Additionally the MSRV is raised to 1.46, since apparently a change at crates.io surfaced an old cargo bug which was fixed in cargo 1.46.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -20,7 +20,7 @@ jobs:
         os:
           - windows-2019 # windows-latest is currently not having a supported MSVC compiler
           - ubuntu-latest
-          - macos-10.15
+          - macos-12
         arch:
           - x86_64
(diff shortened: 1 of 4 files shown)


jschwe:
Looks like there is some issue with crates.io and rust / cargo 1.44.1. I'll try rerunning the pipeline tommorow.

jschwe:
@ogoffart I'm going to go ahead and merge this PR which also raises the MSRV to 1.46 in order to get the CI working again.

PR #194, opened 2022-07-30
Update dependencies

JBBgameich:


generator/Cargo.toml
```diff
@@ -6,13 +6,14 @@ license = "MIT"
...
 [dependencies]
-cargo_metadata = "0.14"
+cargo_metadata = "0.15"
 semver = "1.0"
-platforms = "2.0"
+platforms = "3.0"
 
 [dependencies.clap]
-version = "2.34"
+version = "3.2"
```
jschwe, line 14: Clap 3.2 has an MSRV of 1.56 as far as I can see, while the generator is still supposed to support Rust down to 1.44.1.

jschwe:
Closing, since clap currently can't be updated due to our MSRV policy and the remaining dependencies have either been removed or updated in the meantime.

PR #196, opened 2022-08-06
Add example with cbindgen

### LINKED ISSUE #195: Interoperation with `cbindgen`
I haven't found anything regarding using header files generated from `cbindgen` for shared libraries written in Rust.

I think it might be a nice addition from the point of view of someone like me who has very little experience with CMake.

Would it be ok if I made a PR with a working solution I've come up with and we can iterate on that since I'm sure I've broken some best practices?

mladedav:
/closes #195


jschwe:
Looking at this, it seems we could add a small abstraction for this to corrosion rather easily, since the configuration seems to be done in the `cbindgen.toml` file.

We could add a `generate_bindings(target config_file_path header_name)` function which does something like this:
- call cbindgen and let it generate the headerfile somewhere into the cmake build directory
- add the generated header file to the include directories via `target_include_directories()`

This would require cbindgen to be installed of course. 

Would you be interested in adding something like this?

jschwe:
Closing this one in favor of #330, which adds a helper function to corrosion to generate bindings without needing a build script.

PR #197, opened 2022-08-06
Pass additonal argument to cargo build

### LINKED ISSUE #132: Provide option to pass additional parameters to cargo build
Currently we support a limited subset of arguments that can be passed on to the cargo build invocation.

This could be used to e.g. set `unstable` flags, or currently unsupported stable flags like  `--frozen`/`--locked`/`--offline`/`--message-format`.

mladedav:
/closes #132

I was able to make additional arguments work, but I've seen a lot of things I don't understand the use of (e.g. the whole generator) so some things are probably missing.

Originally I named it FLAGS, but I guess ARGUMENTS, ARGS, or something else would be much more suitable and I'm open to any suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,7 @@ Some configuration options can be specified individually for each target. You ca
   For a given target, enable specific features via `FEATURES`, toggle `ALL_FEATURES` on or off or disable all features
   via `NO_DEFAULT_FEATURES`. For more information on features, please see also the
   [cargo reference](https://doc.rust-lang.org/cargo/reference/features.html).
-- `corrosion_set_flags(<target_name> FLAGS <flag1> ...])`:
+- `corrosion_set_flags(<target_name> <flag1> ...])`:
   For a given target, add options and flags at the end of `cargo build` invocation. This will be appended after any
   arguments passed through the `FLAGS` during the crate import.
 
--- a/README.md
+++ b/README.md
@@ -198,6 +198,8 @@ corrosion_import_crate(MANIFEST_PATH <path/to/cargo.toml>
         [CRATES <crate1> ... <crateN>]
         # Enable the specified features
         [FEATURES <feature1> ... <featureN>]
+        # Pass additional arguments to `cargo build`
+        [FLAGS <flag1> ... <flagN>]
 )
 ```
 
@@ -224,6 +226,9 @@ Some configuration options can be specified individually for each target. You ca
   For a given target, enable specific features via `FEATURES`, toggle `ALL_FEATURES` on or off or disable all features
   via `NO_DEFAULT_FEATURES`. For more information on features, please see also the
   [cargo reference](https://doc.rust-lang.org/cargo/reference/features.html).
+- `corrosion_set_flags(<target_name> FLAGS <flag1> ...])`:
+  For a given target, add options and flags at the end of `cargo build` invocation. This will be appended after any
+  arguments passed through the `FLAGS` during the crate import.
 
 ### Selecting a custom cargo profile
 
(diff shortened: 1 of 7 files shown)


cmake/Corrosion.cmake
```diff
@@ -120,6 +120,7 @@ if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.19.0)
     set(_CORR_PROP_NO_DEFAULT_FEATURES CORROSION_NO_DEFAULT_FEATURES CACHE INTERNAL "")
     set(_CORR_PROP_ENV_VARS CORROSION_ENVIRONMENT_VARIABLES CACHE INTERNAL "")
     set(_CORR_PROP_HOST_BUILD CORROSION_USE_HOST_BUILD CACHE INTERNAL "")
+    set(_CORR_PROP_FLAGS CORROSION_CARGO_FLAGS CACHE INTERNAL "")
```
jschwe, line 123: This if/else block was added for compatibilty reasons to not have to rename existing property names.
For new properties don't extend this if else block. Insteasd just use `INTERFACE_CORROSION_CARGO_FLAGS` as the property name. The `INTERFACE_` prefix is required to support CMake versions before 3.19.

cmake/Corrosion.cmake
```diff
@@ -246,6 +247,8 @@ function(_add_cargo_build)
     set(cargo_target_option "$<IF:${if_not_host_build_condition},--target=${_CORROSION_RUST_CARGO_TARGET},--target=${_CORROSION_RUST_CARGO_HOST_TARGET}>")
     set(target_artifact_dir "$<IF:${if_not_host_build_condition},${_CORROSION_RUST_CARGO_TARGET},${_CORROSION_RUST_CARGO_HOST_TARGET}>")
 
+    set(flags_genex "$<TARGET_PROPERTY:${target_name},${_CORR_PROP_FLAGS}>")
```
jschwe, line 250: I think it would be nice to also support generator expressions for the additional arguments. This is e.g. already done for the `features` genex. Basically wrapping this expression like `$<GENEX_EVAL:$<TARGET_PROPERTY:${target_name},INTERFACE_CORROSION_CARGO_FLAGS>>` should be sufficient.
But maybe you can take a look a the test for `features` and add a similar test with a Genex for the arbitrary flags.

cmake/Corrosion.cmake
```diff
@@ -548,6 +558,26 @@ function(corrosion_set_env_vars target_name env_var)
...
+    set(options)
+    set(one_value_args)
+    set(multi_value_args FLAGS)
+    cmake_parse_arguments(
+            SET
+            "${options}"
+            "${one_value_args}"
+            "${multi_value_args}"
+            "${ARGN}"
+    )
```
jschwe, line 572: Just remove the whole part with `cmake_parse_arguments`.
The signature can just be `corrosion_set_cargo_flags(<target_name> <flag1> [.... <flagN>])`
Below in the set_property call you can just use `${ARGN}` instead of `${SET_FLAGS}`

jschwe (changes requested):
Thanks for your PR! Looks good in general, just some minor things need to be addressed - See my comments.
Also I'd ask you to rebase. I just merged a change which fixes a recent CI issue.

cmake/Corrosion.cmake
mladedav: I've added that now, but I must admit I have no experience with genex. I've practically copied what was in features and then tried to simplify where I could.

README.md
```diff
@@ -224,6 +226,9 @@ Some configuration options can be specified individually for each target. You ca
   For a given target, enable specific features via `FEATURES`, toggle `ALL_FEATURES` on or off or disable all features
   via `NO_DEFAULT_FEATURES`. For more information on features, please see also the
   [cargo reference](https://doc.rust-lang.org/cargo/reference/features.html).
+- `corrosion_set_flags(<target_name> FLAGS <flag1> ...])`:
```
jschwe, line 229:
```suggestion
- `corrosion_set_flags(<target_name> <flag1> ...])`:
```

test/cargo_flags/rust/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "flags-lib"
+version = "0.1.0"
+edition = "2021"
```
jschwe, line 4:
```suggestion
edition = "2018"
```

This should also fix the CI failure

jschwe (approved):
Looks about right now. If you change the edition to 2018 in the test, CI should pass too.

PR #205, opened 2022-08-28
Switch `cargo build` working directory to the Manifest (cargo.toml) directory

### LINKED ISSUE #150: usage of .cargo/config
I was looking to see if it was possible to use a .cargo/config.toml (in my source directory),
I didn't find a way to do so, this line in [FindRust.cmake](https://github.com/corrosion-rs/corrosion/blob/master/cmake/FindRust.cmake#L347-L353) generates a `.cargo/config` in the build directory which is used instead.

This sets the `target-dir` and optionally `rustcflags` fields.
I was midly curious if it would be possible to move this to use the `--target-dir` arg, or `CARGO_TARGET_DIR` or `CARGO_BUILD_TARGET_DIR` env vars instead.

This made me curious what the order of lookups for this setting is, and it appears to be the order in which I have listed them,
i.e. `--target-dir` will take precedence, and all the above will override `.cargo/config`.

as such, the following will fail, 
```
cmake -GNinja ..
CARGO_TARGET_DIR="foo" ninja
```

when "cargo/build" eventually makes it's way to `build_byproducts` in https://github.com/corrosion-rs/corrosion/blob/master/cmake/Corrosion.cmake#L299

as far as I could see, that was the only place its relied upon though.

jschwe:
The output directory (target-dir) is left unchanged.
Changing the cwd to the directory of the Manifest, has the advantage that
cargo can pick up `.cargo/config.toml` files in the source tree.
In the future this would also allow corrosion to respect `rust-toolchain.toml`
and other rustup toolchain overrides.

Closes #150

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 build*/
 install*/
 .vscode
+.idea
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -5,6 +5,11 @@
 - The minimum supported rust version was increased to 1.46, due to a cargo issue that recently
   surfaced on CI when using crates.io.
 
+# Potentially breaking
+
+- The working directory when invoking `cargo build` was changed to the directory of the Manifest
+  file. This now allows cargo to pick up `.cargo/config.toml` files located in the source tree.
+
 # 0.2.1 (2022-05-07)
 
 ## Fixes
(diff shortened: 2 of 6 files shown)


ogoffart (approved):

ogoffart:
The CI is failing though (when Corrosion is used as a library)

jschwe:
> The CI is failing though (when Corrosion is used as a library)

Fixed. The test requires Cargo >= 1.56, since that is when support for the `env` key was added. I had a look at the other available keys in 1.46.0, but nothing really stood out in terms of being suitable for a _simple_ test if the config.toml was read.

PR #227, opened 2022-09-28
Fix target-triple parsing for thumbv8m targets

### LINKED ISSUE #226: Problem with arm/thumb v8
Hi
First, thank you for Corrosion!

The arm/thumb triplet for arm/thumb v8 is a bit special. The available rust toolchains are : 
thumbv8m.base-none-eabi 
thumbv8m.main-none-eabi
thumbv8m.main-none-eabihf

From what i gathered, the dot after v8m is confusing corrosion as it thinks this is a file to load and an internal test fails as it does not end with .json.

The easy workaround is just to remove the test in corrosion.
I don't know how to fix that in a better way, apart from changing the test to check if the target explicitly ends with ".json".

Thank you

PS: This arch is used by Cortex M23 and Cortex M33 MCUs

jschwe:
These targets may contain a dot in the arch field.

Closes #226

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -101,15 +101,9 @@ function(_corrosion_parse_platform manifest rust_version target_triple)
     # except the filename from `target_triple`.
     get_filename_component(target_triple_ext "${target_triple}" EXT)
     if(target_triple_ext)
-        if(NOT (target_triple_ext STREQUAL ".json"))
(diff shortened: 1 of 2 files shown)


jschwe:
@mean00 This fix should solve your issue. Can you confirm?

mean00:
Yes, works fine now

Thank you!

PR #242, opened 2022-10-24
Read OUTPUT_DIRECTORY properties from main library target

### LINKED ISSUE #240: OUTPUT_DIRECTORY and linking against a corrosion-produced library
I was using corrosion (a fairly [recent version](https://github.com/corrosion-rs/corrosion/commit/bff871054636297221c82ccb5647399b2ce4258b)) in a subproject to build a thin wrapper around the oxipng API, and was then trying to dynamically link my main c++ project against it. I was trying to do it this way vs statically linking because my main app and test executables are pretty small, whereas oxipng is pretty large. I set `LIBRARY_OUTPUT_DIRECTORY` to `${CMAKE_BINARY_DIR}/lib` which caused the library to be put in the correct place but produced the `make` error `No rule to make target 'vendor/oxipng/oxipng_encoder.dylib', needed by 'app'. Stop.` (despite the dylib now being in `lib/`)

I created a minimal replication here: https://github.com/corrosion-rs/corrosion/compare/master...mnutt:corrosion:output-directory-linking

which just creates a `consumer` executable and links the corrosion-produced library. It produces the error:

```
[ 50%] Building CXX object CMakeFiles/consumer.dir/consumer.cpp.o
gmake[2]: *** No rule to make target 'librust_lib.a', needed by 'consumer'.  Stop.
gmake[1]: *** [CMakeFiles/Makefile2:374: CMakeFiles/consumer.dir/all] Error 2
gmake: *** [Makefile:91: all] Error 2
CMake Error at ./corrosion/test/ConfigureAndBuild.cmake:103 (message):
  Build step failed.  Exit code: `2`
```

I tried with ninja just to see what would happen and got a very similar error. I observed it on both mac/clang and linux/gcc. Apologies if I'm just using it incorrectly or if what I'm trying to do shouldn't be possible, but I'm guessing the dependency resolution step is somehow happening prior to the output_directory move? FWIW the workaround I eventually used was to statically link the corrosion-produced target against an intermediate `SHARED` target (along with some `-force_load`/`--whole-archive` stuff) and link _that_ to my main app, which worked ok.

jschwe:
Previously, the properties were read from the wrong target. The IMPORTED_LOCATION property needs to be set on the internal target, but the OUTPUT_DIRECTORY properties set by the user will be set on the user facing target (without -static or -shared suffix).

Closes #240

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -123,12 +123,21 @@ endfunction()
 # directory target property value at configure time. This function must be deferred to the end of
 # the configure stage, so we can be sure that the output directory is not modified afterwards.
 function(_corrosion_set_imported_location_deferred target_name base_property output_directory_property filename)
-    get_target_property(output_directory ${target_name} "${output_directory_property}")
+    # The output directory property is expected to be set on the exposed target (without postfix),
+    # but we need to set the imported location on the actual library target with postfix.
+    if("${target_name}" MATCHES "^([^-]+)-(static|shared)$")
+        set(output_dir_prop_target_name "${CMAKE_MATCH_1}")
+    else()
+        set(output_dir_prop_target_name "${target_name}")
+    endif()
+
+    get_target_property(output_directory "${output_dir_prop_target_name}" "${output_directory_property}")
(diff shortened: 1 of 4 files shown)


jschwe:
@mnutt Could you have a look if this PR also fixes your original problem?

Issue #258, opened 2022-11-06
Generated headers should be included via BUILD_INTERFACE / INSTALL_INTERFACE


jschwe:
I have a C++ library that uses a rust library via cxx and corrosion previously.

I tried using the latest commit of corrosion (b921fd09155c05255a6b0df6d9901f919ce81b6b) to use this enhancement to simplify the code, but ran into an issue:
```
CMake Error in CMakeLists.txt:
  Target "my-rust-lib-bridge" INTERFACE_INCLUDE_DIRECTORIES property contains
  path:

    "/Users/ss/projects/my-rust-lib/build-cmake-debug-osx/corrosion_generated/cxxbridge/my-rust-lib-bridge/include"

  which is prefixed in the build directory.


CMake Error in CMakeLists.txt:
  Target "my-rust-lib-bridge" INTERFACE_INCLUDE_DIRECTORIES property contains
  path:

    "/Users/ss/projects/my-rust-lib/build-cmake-debug-osx/corrosion_generated/cxxbridge/my-rust-lib-bridge/include"

  which is prefixed in the build directory.Target "my-rust-lib-bridge"
  INTERFACE_INCLUDE_DIRECTORIES property contains path:

    "/Users/ss/projects/my-rust-lib/build-cmake-debug-osx/corrosion_generated/cxxbridge/my-rust-lib-bridge/include"

  which is prefixed in the source directory.

```

This is probably because I exported my C++ library for external use.

I browsed the source code briefly, is it because https://github.com/corrosion-rs/corrosion/blob/b921fd09155c05255a6b0df6d9901f919ce81b6b/cmake/Corrosion.cmake#L1477
```
target_include_directories(${cxx_target} PUBLIC ${generated_dir}/include)
```

Should we change it to something like below?
```
target_include_directories(${cxx_target}
             PUBLIC
                 $<BUILD_INTERFACE:${generated_dir}/include>
                 $<INSTALL_INTERFACE:include>
             )
```

_Originally posted by @niyue in https://github.com/corrosion-rs/corrosion/issues/244#issuecomment-1304781030_

jschwe:
> Should we change it to something like below?
> 
> ```cmake
> target_include_directories(${cxx_target}
>              PUBLIC
>                  $<BUILD_INTERFACE:${generated_dir}/include>
>                  $<INSTALL_INTERFACE:include>
>              )
> ```

Sounds reasonable to me. Does it fix the CMake error?

[...]

PR #260, opened 2022-11-07
Specify build/install interface for cxx target include directory

niyue:
Specify build/install interface for cxx target include directory to make the include directory is portable.

This PR intends to fix https://github.com/corrosion-rs/corrosion/issues/258


jschwe:
Thanks!

PR #268, opened 2022-11-23
Fix output_directory support when not used as target property

tronical:
In CMake, for example RUNTIME_OUTPUT_DIRECTORY is a target property. It is documented to be initialized from CMAKE_RUNTIME_OUTPUT_DIRECTORY, so it's not unusual for CMake projects to just set the latter variable in order to place all affected targets into the directory, instead of setting a per-target property on each manually.

Since Corrosion respects the target property, it makes sense to assume that just setting the CMAKE_ variable would also work. Oddly, the get_property returns -NOTFOUND when only the CMAKE_ variable is set and not the target property. To "work around" that, this patch implements the fallback manually.


tronical:
I need some advice on how to best add test coverage for this. `test/output directory/output directory/CMakeLists.txt` tests the target properties, but I'd rather not duplicate the whole test. Should I duplicate the series in `test/output directory/CMakeLists.txt` perhaps and add an option for `test/output directory/output directory/CMakeLists.txt` to select whether to set the target property or the global variable?

jschwe:
>  it makes sense to assume that just setting the CMAKE_ variable would also work. Oddly, the get_property returns -NOTFOUND when only the CMAKE_ variable is set and not the target property. 

I guess CMake doesn't set this property on `IMPORTED` or `INTERFACE` targets. I'm not sure why I didn't notice this earlier. Thanks for reporting this!

Instead of complicating the checking part like this PR proposes, I would prefer if Corrosion would initialize the properties as CMake would do.

Since respecting OUTPUT_DIRECTORY anyway requires CMake 3.19, I would suggest editing `_generator_add_cargo_targets()` in `CorrosionGenerator.cmake` and add a loop at the end of the function that iterates over `created_targets` and sets the 5 properties if the relevant `CMAKE_` variable is defined. Should be very simple to do.

This approach has the advantage of working as expected if (for some reason) a user has two or more `corrosion_import_crate` calls in the same CMakeLists.txt file and changes the `CMAKE_` values in between the calls. 

> Should I duplicate the series in test/output directory/CMakeLists.txt perhaps 

Yes, please

> and add an option for test/output directory/output directory/CMakeLists.txt to select whether to set the target property or the global variable?

Thats one option. We could also just duplicate the rust package (and have a CMake test project with 2 corrosion_import_crate calls). In my opinion, less logic in `output directory/output directory` is preferable.

tronical:
Lol, shortly after posting the comment I found the issue. So it's passing locally now.

tronical:
I guess the remaining failures are due to lack of support for `CMAKE_CONFIGURATION_TYPES`. I'll look into that tomorrow.

test/output directory/output directory/CMakeLists.txt
```diff
@@ -2,30 +2,44 @@ cmake_minimum_required(VERSION 3.15)
...
     COMMAND
-    ${CMAKE_COMMAND} -E copy_if_different "$<TARGET_PROPERTY:rust_bin,LOCATION>" "${CMAKE_CURRENT_BINARY_DIR}/another_dir/moved_bin"
+    ${CMAKE_COMMAND} -E copy_if_different "$<TARGET_PROPERTY:rust_bin1,LOCATION>" "${CMAKE_CURRENT_BINARY_DIR}/another_dir/moved_bin"
     )
 
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/custom_bin_var")
+set(CMAKE_PDB_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/custom_bin_pdb_var")
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/custom_archive_var")
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/custom_lib_var")
+set(CMAKE_PDB_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/custom_lib_pdb_var")
```
jschwe, line 32: Currently you are overwriting `CMAKE_PDB_OUTPUT_DIRECTORY` here which is probably not intended. Where you perhaps thinking of `CMAKE_COMPILE_PDB_OUTPUT_DIRECTORY`? I previously ignored the existence of this property, because I was under the impression that Rust does not generate a pdb file for static libraries, but I don't think I double checked that. If rust does generate a pdb file for static libs, then `COMPILE_PDB_OUTPUT_DIRECTORY` is probably the relevant property for that artifact, but I'm honestly not too familiar with windows and implibs.

cmake/CorrosionGenerator.cmake
```diff
@@ -206,4 +206,13 @@ function(_generator_add_cargo_targets)
     else()
         message(DEBUG "Corrosion created the following CMake targets: ${curr_created_targets}")
     endif()
+
+    foreach(target_name ${created_targets})
+        foreach(output_var RUNTIME_OUTPUT_DIRECTORY PDB_OUTPUT_DIRECTORY ARCHIVE_OUTPUT_DIRECTORY LIBRARY_OUTPUT_DIRECTORY PDB_OUTPUT_DIRECTORY)
+            get_target_property(output_dir ${target_name} "${output_var}")
+            if (NOT output_dir AND DEFINED "CMAKE_${output_var}")
+                set_property(TARGET ${target_name} PROPERTY ${output_var} "${CMAKE_${output_var}}")
```
jschwe, line 214: We would probably also want to add a loop which does this for all `CMAKE_CONFIGURATION_TYPES` and the respective `_<CONFIG>` properties.

test/output directory/output directory/CMakeLists.txt
tronical: The use of `CMAKE_PDB_OUTPUT_DIRECTORY` was intentional here, as that would otherwise be used to initialized the `PDB_OUTPUT_DIRECTORY` target property that the test sets further above when not using variables.

I think changing from `PDB_OUTPUT_DIRECTORY` to `COMPILE_PDB_OUTPUT_DIRECTORY` could be done in a separate (unrelated) change. For now I think it makes more sense to keep the symmetry.

cmake/CorrosionGenerator.cmake
```diff
@@ -206,4 +206,13 @@ function(_generator_add_cargo_targets)
     else()
         message(DEBUG "Corrosion created the following CMake targets: ${curr_created_targets}")
     endif()
+
+    foreach(target_name ${created_targets})
+        foreach(output_var RUNTIME_OUTPUT_DIRECTORY PDB_OUTPUT_DIRECTORY ARCHIVE_OUTPUT_DIRECTORY LIBRARY_OUTPUT_DIRECTORY PDB_OUTPUT_DIRECTORY)
```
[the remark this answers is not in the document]
tronical: Oops :). I'll push this and other fixes as follow-up commits and when the change is ready for merging, I'll squash.

cmake/CorrosionGenerator.cmake
tronical: Ah yes! That makes sense and will hopefully fix the test failure :)

test/output directory/output directory/CMakeLists.txt
jschwe: I don't quite follow. From my understanding the assignment in line 32 would overwrite the previous assignment in line 29, so only one of those should be correct.

tronical: Ahh right. I read too fast :(. I think it's actually slightly more complicated. Let's ignore the static library case for a moment:

The target property test places the pdb for the bin into one directory, and the pdb for the (shared) library into another. That is what the target property allows doing. But with the variable based approach both have to end up in the same expected directory, because the same variable feeds them.

tronical:
Ok, at least the CI is green now :-).

jschwe (approved):
Looks good now. Thanks for fixing this! 
You mentioned you wanted to squash, so I'll wait for that before merging.

tronical:
Done :)

tronical:
Hm, for some reason GitHub isn't show that force push.

PR #269, opened 2022-11-26
Support CRATE_TYPES argument for corrosion_import_crate

vdimir:
Consider we have `Cargo.toml` with `crate-type = ["staticlib", "cdylib"]`, but you need to build and use only one of them.
Editing of `Cargo.toml` is not an option (e.g., because it's in a third-party library).
Right now, you can't specify which type to use (adding a such option to `corrosion_import_crate` was my first thought).
But actually, you can set `FLAGS --crate-type=staticlib`, so `*.so` file won't be tried to build. Then, the only setback is `copy_if_different` still would try to copy the file. So, as suggested in this PR, we can probably operate only with files corresponding to the current `BUILD_SHARED_LIBS` settings. 
Any feedback will be appreciated!


jschwe:
Do you have a real-world usecase where the overhead of building the additional lib type is significant and a bottleneck?

vdimir:
In my case, shared lib is failed to build (for some reason, it tries to link with `gss_s` even `NO_STD` is specified, but `libgcc_s.so` is not available in the toolchain, and it also quite difficult to edit). Most likely, it can be managed to build it, but the option to disable unrequited build looks more natural.

jschwe:
The request itself is reasonable, however it would be a breaking change and it is quite likely that there are users relying on cdylibs always being built (thats also why the tests are failing here).

I'll have to think about if we can support this without breaking existing users.

vdimir:
What do you think about the solution with introducing an option to `corrosion_import_crate`, like `corrosion_import_crate(MANIFEST_PATH rust-lib/Cargo.toml CRATE_TYPE staticlib)` ?

jschwe:
> What do you think about the solution with introducing an option to corrosion_import_crate, like corrosion_import_crate(MANIFEST_PATH rust-lib/Cargo.toml CRATE_TYPE staticlib) ?

I guess we could do that. Though I would suggest `CRATE_TYPES` instead and allow the user to specify multiple types (e.g. `staticlib` _and_ `bin` ).

cmake/CorrosionGenerator.cmake
```diff
@@ -52,7 +52,9 @@ function(_generator_add_package_targets workspace_manifest_path package_manifest
         set(kinds)
         foreach(ix RANGE ${target_kind_len-1})
             string(JSON kind GET "${target_kind}" ${ix})
-            list(APPEND kinds ${kind})
+            if(NOT COR_CRATE_TYPES OR ${kind} IN_LIST COR_CRATE_TYPES)
```
jschwe, line 55: `COR_CRATE_TYPES` should be passed as a parameter to this function.
This currently works because in CMake we can also see all variables defined in parent scopes, but we shouldn't rely on this.

test/crate_type/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+corrosion_tests_add_test(crate_type "cpp-exe")
```
jschwe, line 1: It seems like the test is currently not added. You have to add the following in the `tests` directory CMakeLists.txt.
```cmake
add_subdirectory(crate_type)
```
Since `--crate-type=` is used in the test and is a rather new cargo flag, the test also should only be conditionally added if the Rust version is new enough.

jschwe (commented):
Looks mostly good. I left some comments which should be addressed.
Also it would be good if you could add a description of the parameter to the Readme, and also note that the parameter currently only works with CMake >= 3.19.

jschwe (approved):

jschwe:
Looks good to me. Could you rebase and squash?

README.md
```diff
@@ -194,6 +194,8 @@ corrosion_import_crate(MANIFEST_PATH <path/to/cargo.toml>
         [NO_STD]
         # Specify  cargo build profile (e.g. release or a custom profile)
         [PROFILE <cargo-profile>]
+        # Build only the specified crate types
```
jschwe, line 197:
```suggestion
        # Build only the specified crate types (Ignored with CMake < 3.19)
```

PR #272, opened 2022-12-12
Added no linker override flag to corrosion import.

All8Up:
corrosion_import_crate now accepts: NO_LINKER_OVERRIDE
The end result of this is that when it gets time to add the custom build target, 'cargo_target_linker' will be set to an empty string to prevent changing the linker.  This should fix issues where the user has selected the clang compiler and crates such as Winapi that do various things in 'build.rs' would fail due to the incorrect linker being specified for the host.


jschwe:
Could you add a short sentence to the Readme desribing the option?

All8Up:
Err, oops.  Meant to do that.  Will be a bit till I get home.

README.md
```diff
@@ -193,6 +193,8 @@ corrosion_import_crate(MANIFEST_PATH <path/to/cargo.toml>
         # Disable linking of standard libraries (required for no_std crates).
         [NO_STD]
         # Specify  cargo build profile (e.g. release or a custom profile)
+        [NO_LINKER_OVERRIDE]
+        # Disable corrosions normal liker override mechanism.  (Needed occasionally for unique build setups.)
```
jschwe, line 197:
```suggestion
        # Will let Rust/Cargo determine which linker to use instead of corrosion (when linking is invoked by Rust) (Ignored with CMake < 3.19)
```

jschwe (approved):
LGTM. If you fix the conflict, I can merge.

PR #279, opened 2022-12-19
Speedup CI

jschwe:
- Use vendoring to skip the "Updating crates.io index" step, which takes a lot of time. The vendored dependencies are put into a submodule. @ogoffart Do you have permissions to create a new repository in the `corrosion-rs` organization? Currently, I just created a repository for the submodule in my own namespace, but ideally it would be under the organizations' namespace.
- Use `--locked` if a lock-file is present when invoking cargo metadata.


ogoffart:
You should have gotten an invitation to get more right to the corrosion-rs organization.


I'm not a big fan of using a submodule for this purpose.
Can we not reach the same effect by using proper caching in the CI with the cache action?

jschwe:
> You should have gotten an invitation to get more right to the corrosion-rs organization.

Thanks, I received the invitation

> I'm not a big fan of using a submodule for this purpose.

Just to clarify, the submodule would only be used in CI and users would never have to initialize the submodule.

> Can we not reach the same effect by using proper caching in the CI with the cache action?

We could use the cache action to cache the crates-io index, but this would cache the _whole_ index which is huge - That is why it is taking so long to update the index in the first place. I'm not sure if it is possible / sensible to use the action to cache a subset of the index.
For Linux we could use a prebuilt docker image, but that doesn't speed up the macOS and windows jobs.

The vendoring approach has the advantage that it only contains the select few crates we actually need (`cc` for a build script and the dependencies of the `generator` crate), so cloning the submodule is very fast.

In the future this will hopefully be resolved by the sparse checkout over http feature the cargo team is working on, but until this stabilizes and we bump our MSRV it will likely still require a long time.

ogoffart:
If you think it is the best approach, i trust your judgement.

> The vendoring approach has the advantage that it only contains the select few crates we actually need

The cache should normally also only cache the needed crates.

(But it is true that the index is 300M, that's quite big, i hope future version of cargo optimizes that.)

jschwe:
Superseded by #282.

> We could use the cache action to cache the crates-io index, but this would cache the whole index which is huge 

It turns out that using the cache action to restore the cargo registry is _much_ faster than whatever git does, so just caching the whole cache is actually a viable option.

PR #291, opened 2023-01-02
Fix clean target when cross-compiling

tronical:
Use the target parameter for "cargo clean" as for build.


jschwe (approved):
Thanks! looks good.

PR #293, opened 2023-01-05
cxxbridge: Relax build dependency

jschwe:
the cxxbridge command can run before cargo build, so change it to a regular OUTPUT custom command without any
dependency on a target except `cxxbridge`.
This is important for usecases where a C library depends on generated headers, but is later linked into Rust. This means the headers must be built before the C library, and the C library before the Rust target.


Xaldew:
Sorry for the delay, took a bit longer to test this properly than I anticipated.

I applied these changes and tried it on my project, and I noticed a few things, most notably is that I use the [autobins](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#target-auto-discovery) pretty extensively. This yields many warnings of the kind:

```
CMake Warning at /Users/gustaf/.local/share/cmake/CorrosionGenerator.cmake:77 (message):
  Failed to import Rust crate mandelbrot (kind: `[ "bin" ]`) because a target
  with the same name already exists.  Skipping this target.

  Help: If you are importing a package which exposes both a `lib` and a `bin`
  target, please consider explicitly naming the targets in your `Cargo.toml`
  manifest.

  Note: If you have multiple different packages which have targets with the
  same name, please note that this is currently not supported by Corrosion.
  Feel free to open an issue on Github to request supporting this scenario.
Call Stack (most recent call first):
  /Users/gustaf/.local/share/cmake/CorrosionGenerator.cmake:218 (_generator_add_package_targets)
  /Users/gustaf/.local/share/cmake/Corrosion.cmake:1108 (_generator_add_cargo_targets)
  CMakeLists.txt:121 (corrosion_import_crate)
```

There probably needs to be a way to detect this, or at least disable these warnings by the user if this feature is used.

For some reason though, it fails to import the primary library with this error:

```
CMake Warning at /Users/gustaf/.local/share/cmake/CorrosionGenerator.cmake:77 (message):
  Failed to import Rust crate gfx (kind: `[ "lib", "cdylib" ]`) because a
  target with the same name already exists.  Skipping this target.

  Help: If you are importing a package which exposes both a `lib` and a `bin`
  target, please consider explicitly naming the targets in your `Cargo.toml`
  manifest.

  Note: If you have multiple different packages which have targets with the
  same name, please note that this is currently not supported by Corrosion.
  Feel free to open an issue on Github to request supporting this scenario.
Call Stack (most recent call first):
  /Users/gustaf/.local/share/cmake/CorrosionGenerator.cmake:218 (_generator_add_package_targets)
  /Users/gustaf/.local/share/cmake/Corrosion.cmake:1108 (_generator_add_cargo_targets)
  CMakeLists.txt:121 (corrosion_import_crate)

```

This looks weird to me, since there shouldn't be any other targets with the name `gfx` (CMake code used below):

```
find_package(corrosion)
if (corrosion_FOUND)
  corrosion_import_crate(MANIFEST_PATH ${PROJECT_SOURCE_DIR}/ext/gfx-rs/Cargo.toml)
  corrosion_add_cxxbridge(imageio-cxx
    CRATE cxxbridge-crate
    MANIFEST_PATH ${PROJECT_SOURCE_DIR}/ext/gfx-rs
    FILES imageio_cxx.rs)
  corrosion_link_libraries(gfx imageio-cxx ccl_common)
endif()

```

(The absolute path in `test/cxxbridge/cxxbridge_rust2cpp/rust/src/lib.rs` should probably also be fixed somehow, but I think that was mentioned elsewhere)

jschwe:
> The absolute path in ...

This is fixed now in the test.

PR #305, opened 2023-01-14
Allow user to override rustc/cargo per target

jschwe:
This allows e.g. using the nightly toolchain for only one target

These are experimental options for now and could be changed or removed before a stable release


jschwe:
Note: This would not work well together with workspaces / shared dependencies and is blocked until the target directory can be configured (to prevent rebuilding every time due to toolchain mismatch).

PR #310, opened 2023-01-24
Remove linker preference code

jschwe:
Previously, corrosion set the linker to the C++ compiler in order to automatically link in the necessary libraries for C++ code. In practice this seems to be just "c++", so it seems we can also just use the default linker and let Corrosion specify "-lc++" as a linker argument for rust in case the target links in any C++ code.

This is generally preferable and should be more in line with users expect.

This change could potentially have some subtle pitfalls, so someone more familiar with C++ should review it.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -698,7 +698,7 @@ endif()
 #
 # A target may be either a specific bin
 function(_add_cargo_build out_cargo_build_out_dir)
-    set(options NO_LINKER_OVERRIDE)
+    set(options "")
     set(one_value_args PACKAGE TARGET MANIFEST_PATH PROFILE WORKSPACE_MANIFEST_PATH)
     set(multi_value_args BYPRODUCTS TARGET_KINDS)
     cmake_parse_arguments(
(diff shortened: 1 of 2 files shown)


jschwe:
@ogoffart I'd like to remove the corrosion code that explicitly sets the C++ compiler as the linker driver when Rust is used to invoke the linker and instead specify the libraries that C++ requires (i.e. add `-Clink-arg=-lc++`), so that linking works with any linker. This seems to work just fine on mac - I haven't had a chance yet to look into Linux. 

I don't really have any C++ experience, so I'd like some feedback from C++ developers if this approach is even feasible, or if there are some hidden subtleties.

In any case it would still be possible for the user to explicitly specify a linker - it's just that corrosion would not automatically specify the linker rust should use anymore.

jschwe:
This appears to be a breaking change in regards to cross-compiling "just working" with corrosion for rust targets with a default linker that is not rust-lld or another Cross-linker. It seems that rust doesn't have a default linker for the failing targets we tests, and just defaults to cc, which can't work for cross-compiling if cc is GCC. For clang it would probably also not work, since likely no --target would be specified. For these reasons I'm delaying this change until the affected tool chains link with rust-lld by default, which may still take a while.

tronical:
I think the origin of this was likely something else (“preference”), but I agree it’s crucial for cross compilation. Perhaps the code could be reduced and simplified in that direction (for example replace list with single linker item)

jschwe:
Superseded b #361.

Issue #218, opened 2022-09-19
Add parameter to corrosion_import_crate to return list of added targets.


jschwe:
This could be used by the caller to set some properties for all (or a filtered subset) of targets.
You can already manually set the properties, but in cases where there are a lot of targets, it may be more convenient to just iterate over a list that corrosion provides.

temeddix:
Yeah, this is definitely needed in actual developement.

jschwe:
>  this is definitely needed in actual developement.

I'm curious what your usecase is that you "definitely need" this. From my point of view it's just a minor convenience improvement. Could you elaborate?

temeddix:
We have a situation where we have to compile multiple crates inside a workspace crate. Since our repository is public, it is https://github.com/cunarist/app-template. It's a template for app development written in Fltuter and Rust

We have a workspace rust crate `native` and some other library crates insied it, like this:
- `native`: workspace crate
  -  `first_crate`: library crate
  - `second_crate`: library crate
  - ...(More library crates will be added in the future)

With this situation, we can compile multiple library crates at once with a CMAKE file like this:
```cmake
# Part of CMAKE file
corrosion_import_crate(MANIFEST_PATH ../native/Cargo.toml)
set(CRATE_NAME "first_crate")
target_link_libraries(${BINARY_NAME} PRIVATE ${CRATE_NAME})
list(APPEND PLUGIN_BUNDLED_LIBRARIES $<TARGET_FILE:${CRATE_NAME}-shared>)
set(CRATE_NAME "second_crate")
target_link_libraries(${BINARY_NAME} PRIVATE ${CRATE_NAME})
list(APPEND PLUGIN_BUNDLED_LIBRARIES $<TARGET_FILE:${CRATE_NAME}-shared>)
```
However, if we add new library crate insied the `native` workspace crate, we have to modify this file again and again. Therefore it would be nice to have something like this:
```cmake
# Part of CMAKE file
corrosion_import_crate(MANIFEST_PATH ../native/Cargo.toml)
foreach(LIB ${LIBS}) # Assuming that a list name LIBS is from corrosion_import_crate
if (TARGET ${LIB})
    list(FIND VISITED_TARGETS ${LIB} VISITED)
    if (${VISITED} EQUAL -1)
        get_target_property(LIB_FILE ${LIB} LOCATION)
        get_link_libraries(LINK_LIB_FILES ${LIB})
[...]

PR #312, opened 2023-01-25
Add parameter to corrosion_import_crate to return list of added targets

temeddix:
This PR adds new `IMPORTED_CRATES` flag to `corrosion_import_crate()` to retrieve the list of imported crates in the current scope. Idea was from issue #218.

With this PR, this kind of operation is made possible.
```cmake
# Part of CMAKE file
corrosion_import_crate(MANIFEST_PATH ../native/Cargo.toml IMPORTED_CRATES crates_list)
target_link_libraries(${BINARY_NAME} PRIVATE ${crates_list})
```

I checked that this works as intended in an actual Rust/C++ project.


jschwe:
Thanks for the PR!

PR #317, opened 2023-01-30
Fix links to rust2cpp code in README

redstrate:
The directories got moved at some point, so the links broke in the README.


jschwe:
Thanks 👍

PR #327, opened 2023-02-17
Migrate documentation to mdbook

jschwe:
The Readme has been getting larger and larger. Since there are a lot of topics which could be
added to corrosions documentation, move the documentation to the mdbook format.
This allows structuring the content better in the future.


.github/workflows/gh-pages.yaml
```diff
@@ -0,0 +1,63 @@
+name: Deploy GH pages
+on:
+  push:
+    branches:
+      - master
+      # Todo: Only for testing purposes before merging to master
+      - book
```
jschwe, line 7: Remove me before merging

PR #337, opened 2023-02-28
Improve argument checking and set IMPORTED_NO_SONAME

jschwe:
- Currently imported shared libraries don't have an soname set by default, so we should set `IMPORTED_NO_SONAME`.
- Refactor some of the argument passing, so we can have better argument checking. Mostly important to ensure that arguments are actually passed around and not just happen to be in scope because the caller had the variable defined too. Also makes calling functions less error prone, since named arguments are better readable than a long list of positional arguments.


jschwe:
bors retry

PR #346, opened 2023-03-03
Add "alpine" as a known triple vendor

ptrcnull:
Alpine uses a separate vendor for Rust installed from aports: `$CARCH-alpine-linux-musl`

See https://gitlab.alpinelinux.org/alpine/aports/-/blob/329ba719/community/rust/alpine-target.patch


jschwe:
Thanks.

PR #353, opened 2023-03-06
If Rust_CARGO_TARGET has been defined, make sure it is used everywhere

mqudsi:
Some cargo/rustc options require an explicit target triple or else compilation will fail. If the user has set Rust_CARGO_TARGET, respect it and use it everywhere.

(Context: building w/ RUSTFLAGS='-Zsanitizer=address' set requires invoking cargo w/ `--target <triple>` for compilation to succeed or else linker errors will block compilation from completing successfully.)

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -677,6 +677,9 @@ if(CORROSION_NATIVE_TOOLING)
         if(NOT _CORROSION_VERBOSE_OUTPUT_FLAG)
             set(generator_build_quiet "--quiet")
         endif()
+        if(Rust_CARGO_TARGET)
+            set(cargo_target_option "--target" ${Rust_CARGO_TARGET})
+        endif()
         # Using cargo install has the advantage of caching the build in the user .cargo directory,
         # so likely the rebuild will be very cheap even after deleting the build directory.
         execute_process(
@@ -684,6 +687,7 @@ if(CORROSION_NATIVE_TOOLING)
                     --path "."
                     --root "${generator_destination}"
                     ${generator_build_quiet}
+                    ${cargo_target_option}
                 WORKING_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}/../generator"
                 RESULT_VARIABLE generator_build_failed
         )


jschwe:
Sorry, but this PR doesn't make any sense. 
You are modifying the target triple for the Rust program which generates some CMake code to import rust crates for CMake versions older than 3.19. 
There is no reason to compile this internal corrosion target with a sanitizer, and especially no reason to potentially cross-compile it, since it should always be compiled for the host.

For the actually imported Rust crates, `--target` is always passed.

mqudsi:
Thanks for explaining. The issue I ran into is that setting RUSTFLAGS is typically done in the environment outside of/before CMake so the helper crate fails to link with the sanitizer set. The only other option I can see here would be requiring cmake to be aware of ASAN and internally use `corrosion_add_target_local_rustflags(target_name "-Zsanitize=memory")`

mqudsi:
Would you be open to a PR that either unset RUSTFLAGS when invoking `cargo` to build corrosion's own crate or (more error-prone) removed any `-Zsanitizer=xxx` flags in it (again for the compilation of its own crate)?

jschwe:
>  in the environment outside of/before CMake 

So in your case CMake is not actually the main buildsystem?

> The only other option I can see here would be requiring cmake to be aware of ASAN 

That is what I would consider the standard approach. Add some CMake `option` to enable ASAN to your project, and depending on that option use `corrosion_add_target_rustflags(target_name "-Zsanitize=address")`. Note that you probably don't want the `local` variant, since probably you want the address sanitizer to be enabled for everything and not just the top-level crate.

> Would you be open to a PR that either unset RUSTFLAGS when invoking cargo to build corrosion's own crate

I would be open to that. Prepending the command with `${CMAKE_COMMAND} -E env --unset=RUSTFLAGS` should not break anything. 

FYI, if you are using CMake < 3.19 it is generally recommended to install corrosion, since that means you only build the Generator once. Afterwards when you use corrosion in your own project the generator would not have to be built, since corrosion uses the installed executable.

mqudsi:
> So in your case CMake is not actually the main buildsystem?

I have a make file that glues things together but the actual case I had in mind is CI where per-target configuration is often done via a configuration file. For the C++ portion of our codebase, we just globally define `CXXFLAGS='-fsanitize=address -fno-omit-frame-pointer'` without needing to have any ASAN-specific code in CMake. I was hoping to do the same w/ our now-hybrid C++/rust codebase.

> Note that you probably don't want the local variant, since probably you want the address sanitizer to be enabled for everything and not just the top-level crate.

Thanks, noted.

> I would be open to that. Prepending the command with ${CMAKE_COMMAND} -E env --unset=RUSTFLAGS should not break anything.

Yes, [I tried it](https://github.com/corrosion-rs/corrosion/commit/fadf4dd4c39951b147107803358640b9b871a102) with the "uglier/unixier" syntax `${CMAKE_COMMAND} -E env RUSTFLAGS= ...` and it worked.

> FYI, if you are using CMake < 3.19 it is generally recommended to install corrosion, since that means you only build the Generator once. Afterwards when you use corrosion in your own project the generator would not have to be built, since corrosion uses the installed executable. Edit: And for CMake >= 3.19 you also don't need the generator, so you don't have this problem either.

Thanks, I didn't know that. I'm on 3.18 so this would save me a lot of time each time I `make clean` and `make` again to test the changes to the build system.

mqudsi:
@jschwe what are the conditions for the locally installed `corrosion-generator` crate to be used? I did `cargo install --path .` in the corrosion generator directory but CMake still fetches and builds corrosion (for a clean build) (cmake 3.18).

jschwe:
@mqudsi Please see the documentation here: https://corrosion-rs.github.io/corrosion/setup_corrosion.html#installation

PR #362, opened 2023-03-10
Improve required library detection for staticlibs

jschwe:
Instead of hand maintaining a list of required libraries, it is more efficient to just ask `rustc` what libraries are needed to link.

Using a demo project will not give a full list, but at least a minimum list of required dependencies - which is the same situation as now.

In the future we could expand this by passing arguments from the cargo build step to the link-step via a file.
But that needs more investigation, so we do this improvement first.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -17,11 +17,23 @@
   default they won't have an `soname` field.
   If you add a rustflag like `-Clink-arg=-Wl,-soname,libmycrate.so` in your project,
   you should set this property to false on the shared rust library.
+- Corrosion now uses a mechanism to determine which native libraries need to be linked with
+  Rust `staticlib` targets into C/C++ targets. The previous mechanism contained a hardcoded list.
+  The new mechanism asks `rustc` which libraries are needed at minimum for a given
+  target triple (with `std` support). This should not be a breaking change, but if you
+  do encounter a new linking issue when upgrading with `staticlib` targets, please open an
+  issue.
 
 ## New features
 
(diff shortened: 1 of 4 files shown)


RELEASES.md
```diff
@@ -17,11 +17,23 @@
...
 ## New features
 
 - `corrosion_import_crate()` has two new options `LOCKED` and `FROZEN` which pass the 
   `--locked` and `--frozen` flags to all invocations of cargo. Only with CMake >= 3.19.
+- `FindRust` now provides cache variables containing information on the default host
+  target triple:
+  - `Rust_CARGO_HOST_TARGET_ARCH`
+  - `Rust_CARGO_HOST_TARGET_VENDOR`
+  - `Rust_CARGO_HOST_TARGET_OS`
+  - `Rust_CARGO_HOST_TARGET_ENV`
```
ogoffart, line 36: Isn't the `HOST_TARGET` a bit strange?  Should that be named `Rust_CARGO_HOST_ARCH` instead (without the TARGET)

jschwe: makes sense, made an issue so I don't forget.

Be-ing:
This seems to have been a helpful change. Somehow we started getting link errors for symbols in NtDll.dll on Windows with GitHub Actions yesterday and updating to Corrosion 0.4.0 fixed those errors.

jschwe:
did you update to rust 1.70 yesterday?

Be-ing:
Yes, we're using https://github.com/actions-rs/toolchain which automatically uses the latest stable.

PR #371, opened 2023-03-27
[Feature] Add parameter for defining native libs to link

dcvz:
Adds a `NATIVE_LIBS` parameter to `corrosion_import_crate` at least while the efforts to better be able to detect and link necessary libs falls in.

There is documentation to update and lose ends to tie, but first wanted to see if this is of interest.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -445,6 +445,10 @@ function(_corrosion_add_library_target)
 
         # Todo: NO_STD target property?
         if(NOT COR_NO_STD)
+        foreach(lib ${COR_NATIVE_LIBS})
+            list(APPEND Rust_CARGO_TARGET_LINK_NATIVE_LIBS ${lib})
+        endforeach()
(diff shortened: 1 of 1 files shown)


jschwe:
Is `corrosion_link_libraries()` not sufficient for your usecase?

jschwe:
> I think because cmake will link them after crate is built?

Ah  - you are building a Rust staticlib correct? In that case, have you tried `target_link_libraries()`?

jschwe:
What exactly did you try? In the end, what you need to do is tell CMake which libraries it should link into your final binary.
Doing `target_link_libraries($your_rust_static_lib_crate_name INTERFACE $your_required_native_libraries )` should do exactly that. 
This will add the libraries to `INTERFACE_LINK_LIBRARIES` of your rust target - the same as what you are proposing to do via a new parameter to `corrosion_import_crate`.

jschwe:
Closing due to lack of response.

PR #373, opened 2023-03-29
Fix cargo tree to use the correct cargo executable

yuval-nextsilicon:
This fixes an issue where using `corrosion_add_cxxbridge` on a system that doesn't have cargo in PATH fails with "Crate ... does not depend on cxx"


jschwe:
Thanks!

Issue #378, opened 2023-04-13
VS platform detection is case sensitive (handles "Win32" but not "win32")


CouleeApps:
The windows platform detection code in FindRust.cmake checks for the platform to be named "Win32" case sensitively, but the default VSCode CMake configuration for amd64_x86 (host x64, target x86) uses `cmake ... -A win32` (lowercase). This causes corrosion-rs to throw an error since it was expecting Win32 (capitalized).

https://github.com/corrosion-rs/corrosion/blob/master/cmake/FindRust.cmake#L583-L593

Proposed fix: use `string(TOLOWER ...)` on the platform name for case-insensitive comparison.

CouleeApps:
Upon further inspection, this might also be a bug in the official VSCode CMake plugin-- every other source I've seen uses capitalized Win32.

PR #379, opened 2023-04-13
Compare VS platform name case insensitively

CouleeApps:
Fixes https://github.com/corrosion-rs/corrosion/issues/378 via the proposed fix.


CouleeApps:
The exact command that VSCode was trying to run (and I have been trying to run from powershell as well) is `cmake -G "Visual Studio 17 2022" -T host=x64 -A win32` which does seem to compile properly with this patch. I have not tried arm64, but I could install the toolchain and give it a try.

jschwe:
>  I have not tried arm64, but I could install the toolchain and give it a try.

That would be helpful.

FYI, the CI failure on windows with rust nightly is a known issue and unrelated to your changes.

jschwe (approved):
I tested and configuring with  `-A arm64` or `-AaRM64` works for CMake, so doing a case insenstive compare makes sense to me.

PR #382, opened 2023-04-15
Improve error message if no package is imported

### LINKED ISSUE #380: Provide a nicer error when no packages are found
It would be nice to provide a better error for a user that might have failed to match CRATES name with the one actually in Cargo.toml
https://github.com/corrosion-rs/corrosion/blob/4c36abff079364d6ebdf51526fc3457c216a0545/cmake/CorrosionGenerator.cmake#L278.

What do you think about if `CRATES` is set by user, variant message listing all packages that have been found?
Something like `Fail to find your-crate,other-crate in N packages (package1,package2) etc...`.

Idea is coming from this user report that had error unrelated to cxx-qt: https://github.com/KDAB/cxx-qt/issues/516

jschwe:
Add some information to the error message on common causes, a list of all packages there were found and print
the `CRATES` whitelist if it set.
It's a bit unfortunate that the CRATES whitelist is actually a whitelist for package names, but changing this is a
breaking change.

Closes #380

### DIFF
--- a/cmake/CorrosionGenerator.cmake
+++ b/cmake/CorrosionGenerator.cmake
@@ -222,18 +222,17 @@ function(_generator_add_cargo_targets)
     math(EXPR ws_mems_len-1 "${ws_mems_len} - 1")
 
     set(created_targets "")
+    set(available_package_names "")
     foreach(ix RANGE ${pkgs_len-1})
         string(JSON pkg GET "${packages}" ${ix})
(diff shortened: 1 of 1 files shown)


jschwe:
CC @OlivierLDff. What do you think about this error message?

OlivierLDff:
Yes this is really nice!

PR #383, opened 2023-04-15
Update native link library list

jschwe:
Update the list based on obvservations from #362.
This is only an intermediate step, since in the future it would be preferable to use an approach based on #362.
However, for now it is easier to update the
hardcoded list.
For windows, this update seems to be necessary for Rust 1.70.


jschwe:
doesn't seem to fix the windows nightly ci failure, so closing.

PR #387, opened 2023-04-24
Fix unnecessary recompilations

### LINKED ISSUE #386: Projects with multiple Rust targets have broken incremental compilation
After updating Corrosion from 0.3.0 to 0.3.5, I noticed every time I build my whole project, it rebuilds all Rust crates including dependencies, instead of building the project incrementally as before. When building the same Rust target twice, the second time does not cause a rebuild (as expected). When building one Rust target, then another, then the first one again, all three times all dependencies are rebuilt from scratch. This makes me think that building a target causes incremental builds to get invalidated for all the other targets.

I tracked it down to commit 798fc952a7753fa8d592bc9f7e37bfe0fe4898fd (introduced between 0.3.1 and 0.3.2), reverting the commit [in my project](https://git.dblsaiko.net/nucom/commit/?h=idl&id=3d3bbcdc63d85c2695179295ef6017fa3277a1d8) fixes the issue.

jschwe:
Changing / unsetting `CARGO_TARGET_xyz_LINKER` causes cargo to recompile all dependencies, which is unnecessary.
To avoid this we now set the linker directly for the last `rustc` invocation. This has the side-effect of not affecting the linker for potential compilations with `cc` in downstream build scripts, which should be an improvement.

We also now pass `linker-flavor` as an argument, which fixes some linker inference issues in older rust versions.

Fixes #386

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -826,10 +826,17 @@ function(_add_cargo_build out_cargo_build_out_dir)
     if(NOT (Rust_CARGO_TARGET_ENV STREQUAL "msvc" OR COR_NO_LINKER_OVERRIDE))
         set(default_linker "$<IF:$<BOOL:${target_uses_cxx}>,${CMAKE_CXX_COMPILER},${CMAKE_C_COMPILER}>")
     endif()
-    # Used to set a linker for a specific target-triple.
-    set(cargo_target_linker_var "CARGO_TARGET_${_CORROSION_RUST_CARGO_TARGET_UPPER}_LINKER")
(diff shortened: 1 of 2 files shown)


jschwe:
This MR has the downside of breaking builds of `bin` crates, that depend on a lib which also has `dylib` or `cdlib` in kinds, and links in C++ code, since the linker is only set by corrosion for the top-level target. 
This probably rules out merging this MR, unless we get some upstream support to only build the `rlib` targets of dependencies.  Not sure if this is possible, since there are flags like `--prefer-dynamic`, which probably require `dylib`s to be built alongside.

PR #391, opened 2023-05-18
Fix determining required link libraries

crabtw:
If binary directory is in workspace, printing libraries will fail.
Adding `[workspace]` to Cargo.toml fixes that besides this patch also fixes a typo.


jschwe:
Looks good, thanks!

Issue #58, opened 2020-10-06
Make imported names more specific


AndrewGaspar:
Currently we only import the primary target of the crate, ignoring the fact that a crate can export multiple bins and libs. So we should really import the primary target as `foo::foo`, and then import alternate targets as, for example, `foo::bar`. Related to #57, #11, #12

jschwe:
Update: We now import all targets, but there is no namespacing happening so multiple targets from different packages may not have overlapping names. Adding a prefix could solve this, but would also be a breaking change.

PR #394, opened 2023-05-26
Added ability to import a crate containing lib and bin targets with the same name

obsgolem:
We do this by adding the ability to put the bin and lib targets into separate CMake namespaces. This also takes some initial steps towards breaking the connection between Rust target names and CMake target names. This structure could be useful for #58 eventually.


jschwe:
I probably won't have time to review this until one week from now.
I'd also be interested if you have a specific usecase where you really need this feature and can't just edit the cargo.toml to get distinct names. This MR does complicate the code and create a maintenance burden, so I'm a bit hesitant.

obsgolem:
So, a decent chunk of this code is just plumbing that would be needed for #58. In fact, I could do up a follow up that implements that issue using a very small amount of code. Since it is shared work, my opinion is that the maintenance burden isn't too bad.

Personally, I could rename my targets in my Rust project, but it really doesn't make sense to. If I have a binary named `x` I should be able to refer to the library code as `use x::whatever` instead of `use libx::whatever`.

obsgolem:
One question: is `IMPORTED_CRATES` intended to yield a list of targets instead of crates? If so, would it be reasonable to add a new `IMPORTED_TARGETS` input and deprecate `IMPORTED_CRATES`?

jschwe:
>  is IMPORTED_CRATES intended to yield a list of targets instead of crates

Well, previously the crate name was the same as the target name (a single rust Cargo.toml **package** manifest contains zero or one library crates, and 0-n bin **crates**).

I'm conservative regarding deprecating / renaming things - I think it should be sufficient to document that the variable will contain a list of CMake targets, representing the crates that were imported. 
I am considering splitting `corrosion_import_crate` into `corrosion_import_workspace` and `corrosion_import_package`, and if I ever do that, then the parameter names for the new functions could also be adjusted.

> I currently use __ when :: is not valid in a target name. 

I would document that the transformation is [`string(MAKE_C_IDENTIFIER)`](https://cmake.org/cmake/help/latest/command/string.html#make-c-identifier), which effectively does the same.

obsgolem:
> [string(MAKE_C_IDENTIFIER)](https://cmake.org/cmake/help/latest/command/string.html#make-c-identifier)
Just to be clear, do you want me to replace my existing `string(REPLACE)` with this?

jschwe:
Yes, please

obsgolem:
Are you sure this has the correct semantics? It converts e.g. `flags-lib` into `flags_lib`. This is currently causing build failures; I can work around it, but personally I prefer the more targeted `string(REPLACE)`.

jschwe:
>  It converts e.g. flags-lib into flags_lib

Ah, I see. No that is not intended. In that case lets stick with the `string(REPLACE)`. Sorry for confusion.

cmake/CorrosionGenerator.cmake
```diff
@@ -111,6 +119,7 @@ function(_generator_add_package_targets)
             _corrosion_add_library_target(
                 WORKSPACE_MANIFEST_PATH "${workspace_manifest_path}"
                 TARGET_NAME "${target_name}"
```
jschwe, line 121: I'm wondering: Is the original `target_name` even still used after we defined `target_name_cmake`? Is there a reason why we would need both target_name and target_name_cmake?

cmake/Corrosion.cmake
```diff
@@ -436,71 +443,73 @@ function(_corrosion_add_library_target)
     endif()
     set("${CALT_OUT_ARCHIVE_OUTPUT_BYPRODUCTS}" "${archive_output_byproducts}" PARENT_SCOPE)
 
-    add_library(${target_name} INTERFACE)
+    add_library(${target_name_cmake} INTERFACE IMPORTED GLOBAL)
```
jschwe, line 446: Is there a specific reason why you added `IMPORTED GLOBAL` here? I'm pretty sure it _probably_ doesn't change anything, but then again this is the kind of change that might subtly break something some user is relying on. I would prefer if this change was removed from this PR. If you would like to add `IMPORTED_GLOBAL`, then please open a separate PR with just this change.

jschwe (approved):
Looks good besides the two comments.
Also it would be good if you could document the two new flags, by adding it to the documentation comment here: https://github.com/corrosion-rs/corrosion/blob/4a134dffa8a9729b7046fe1724bbb00d6999c38d/cmake/Corrosion.cmake#L938
  The documentation will automatically be included into the usage  documentation here: https://corrosion-rs.github.io/corrosion/usage.html

cmake/CorrosionGenerator.cmake
obsgolem: Yes, it is used to form the lib file names.

obsgolem:
Whoo boy, sorry for not getting back to this for so long. I redid the entire MR since it was so far behind and I had forgotten exactly what I had done in the first place.

jschwe:
I'll be travelling a lot in the next 3 weeks, so it might take a while till I can review this PR (again).

cmake/CorrosionGenerator.cmake
```diff
@@ -134,16 +145,18 @@ function(_generator_add_package_targets)
...
             string(REPLACE "\-" "_" target_name "${target_name}")
+            string(REPLACE "\-" "_" target_name_cmake "${target_name_cmake}")
 
             set(archive_byproducts "")
             set(shared_lib_byproduct "")
             set(pdb_byproduct "")
 
-            add_library(${target_name} INTERFACE)
-            _corrosion_initialize_properties(${target_name})
+            add_library(${target_name_cmake} INTERFACE IMPORTED GLOBAL)
```
jschwe, line 154: Whats the reason to make the interface library `IMPORTED` and `GLOBAL`? It doesn't seem directly related to this PR.

cmake/Corrosion.cmake
```diff
@@ -165,6 +166,7 @@ endfunction()
 #    artifact.
 # - filename of the artifact.
 function(_corrosion_set_imported_location target_name base_property output_directory_property filename)
+        message(STATUS ${target_name})
```
jschwe, line 169: probably this debug print should be removed again

cmake/Corrosion.cmake
```diff
@@ -388,79 +395,81 @@ function(_corrosion_add_library_target)
     set("${CALT_OUT_ARCHIVE_OUTPUT_BYPRODUCTS}" "${archive_output_byproducts}" PARENT_SCOPE)
 
     if(has_staticlib)
-        add_library(${target_name}-static STATIC IMPORTED GLOBAL)
-        add_dependencies(${target_name}-static cargo-build_${target_name})
-        set_target_properties(${target_name}-static PROPERTIES COR_FILE_NAME ${static_lib_name})
+        add_library(${target_name_cmake_underscore}-static STATIC IMPORTED GLOBAL)
```
jschwe, line 398: Is there a reason we don't use `target_name_cmake` / `::` for the internal library name?

jschwe (commented):
Left some minor comments. Also for this PR it would be nice if you could add a short description to the release notes (RELEASES.md)

obsgolem:
Is there any chance this can get looked at again soon?

jschwe:
Sorry this fell off my radar. Issue https://github.com/corrosion-rs/corrosion/issues/611 is quite similar. I'm wondering if we can resolve both issues with one solution, so perhaps we can have a short discussion in the issue where I outlined some options that came to my mind.

jschwe:
@obsgolem Could you rebase this PR? (If possible dropping the lockfile changes)

PR #396, opened 2023-05-27
Prepare v0.4 release

jschwe:


RELEASES.md
```diff
@@ -1,8 +1,22 @@
...
 
-Changes compared to v0.3.5:
+No changes compared to v0.4.0-beta2.
 
-## Breaking Changes
+## Announcements
+
+The `v0.4.x` LTS series will be the last release to support older CMake and Rust versions.
+If necessary, fixes will be backported to the v0.4 branch. New features will not be
+actively backported after the v0.5 release, but community contributions are possible.
```
ogoffart, line 9: idea: don't commit the next release v0.5
Maybe it would be time to move to 1.0 at some point?
But that's just details since we can still choose any version number regardless of what's written there.

RELEASES.md
```diff
@@ -1,8 +1,22 @@
...
+The `v0.4.x` series is currently planned to be maintained until the end of 2024.
```
ogoffart, line 10: I'm glad that you are willing to maintain this. If you think you can do it. 
I would be careful to promise anything though, that you may regret later. That said, A line in the ChangeLog doesn't make anyone liable for anything so it's fine to communicate your intend even if we stop maintaining this branch earlier.  I personally would leave the exact date out. Up to you

RELEASES.md
```diff
@@ -1,8 +1,22 @@
-# 0.4.0-beta1 (2023-05-10)
+# 0.4.0 LTS (2023-05-28)
```
ogoffart, line 1: The exact date needs just to be updated

ogoffart (approved):
🥳

(Thank you again for all your work on corrosion!)

RELEASES.md
jschwe: I would say I was quite careful with what I promised: Fixing bugs in the LTS branch and reviewing pull-requests for feature backports. The end of 2024 is in 1.5 years, so the timeframe I promised is also not overly long.

RELEASES.md
ogoffart: I'd say if you know that the next version will have breaking change, then continue with the 0.x series,  if not try to go with 1.x

We could have 1.0 LTS now, then 1.1  (non-lts)  that require new cmake and rust (because we could say that new version of cmake and rust are not breaking change)

jschwe: > I'd say if you know that the next version will have breaking change,

In that case I'll stick with 0.4 for now.

PR #397, opened 2023-05-28
Fix cargo tree to use the correct rustc executable

yuval-nextsilicon:
Sorry, I added this code (in https://github.com/corrosion-rs/corrosion/pull/373) in order to fix an error with `cargo` not being found when it is not in path, but it seems that my patch was not good enough as it now fails because cargo can't find `rustc`... I think this patch makes it work (I can only test this locally because the github CI seems to have the Rust installation configured in PATH)


jschwe (approved):
> I think this patch makes it work

I think so, too. Invoking cargo here is now the same as for the main `cargo build` command.

PR #408, opened 2023-06-18
Install selected toolchain via rustup

yuval-nextsilicon:
Hey :)

I'm not sure if this is the correct way to implement this. I wrote this to bridge a gap in a project that uses `rust-toolchain.toml` to select a toolchain (and possibly download it). It seems that corrosion does support choosing the toolchain that is specified in `rust-toolchain.toml` (because it appears as an override toolchain in rustup and corrosion sets the `_TOOLCHAIN_OVERRIDE` setting), but only if that toolchain is installed, contrary to cargo which downloads the toolchain automatically.

There seems to be a fundamental issue here where rustup doesn't show a toolchain override from `rust-toolchain.toml` until cargo (or something else) ran and installed it, and there doesn't seem to be any rustup subcommand to install whatever toolchain is required by the current project. I thought the corrosion-specific toolchain override is the place to bridge that gap until rustup gains the ability to bootstrap the correct toolchain as the override settings dictate.

### DIFF
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -221,6 +221,7 @@ if (DEFINED Rust_TOOLCHAIN)
     # If the user specifies `Rust_TOOLCHAIN`, then look for `rustup` first, rather than `rustc`.
     find_program(Rust_RUSTUP rustup PATHS "$ENV{HOME}/.cargo/bin")
     if(Rust_RUSTUP)
+        execute_process(COMMAND "${Rust_RUSTUP}" install "${Rust_TOOLCHAIN}")
         set(_RESOLVE_RUSTUP_TOOLCHAINS ON)
     else()
         set(_RESOLVE_RUSTUP_TOOLCHAINS OFF)


jschwe:
Currently corrosion has a lot of complicated logic in `FindRust.cmake` which selects an **actual** rust toolchain, based on e.g. the `Version` criteria a user has set. Corrosion will directly use this toolchain, bypassing the rustup proxy. This is a bit fast when building, since the rustup invocation is skipped, and I would say also has less potential for surprises.

Background: The `cargo` and `rustc` that you have in `.cargo/bin` are actually just `rustup`. rustup will select and potentially install the selected toolchain before then executing cargo or rustc from that toolchain. This means additional overhead on every cargo command, since first rustup will be invoked and check all the potential override sources.

I do agree that having automatic installing of toolchains can be useful, but if you e.g. just want to use the rust version specified in your rust-toolchain.toml, then I would suggest adding an option to FindRust.cmake which skips all the complicated version resolution logic, and will just take the `cargo` / `rustc` that is in `PATH` (which would be the rustup proxy, which will automatically install toolchains ).

yuval-nextsilicon:
I see. How do you think I should implement that, as new CMake variable or some a magic value of `Rust_TOOLCHAIN`?

yuval-nextsilicon:
Also, this proxy behavior of rustup acting is cargo/rustc is pretty funny, makes it even weirder that rustup doesn't support installing the override toolchain when it's acting as rustup

jschwe:
>I see. How do you think I should implement that, as new CMake variable or some a magic value of Rust_TOOLCHAIN?

I would add a new `option()` for this. With `rustup toolchain link` developers can add custom toolchains with arbitrary names, so I don't think a magic value would be good.

> Also, this proxy behavior of rustup acting as cargo/rustc is pretty funny, makes it even weirder that rustup doesn't support installing the override toolchain when it's acting as rustup

Corrosion bypasses the rustup proxy and directly calls the real cargo / rustc, so that is expected.

yuval-nextsilicon:
I'll open it under a different PR

PR #430, opened 2023-08-05
Fix Legacy CMake Generator with nightly Rust

jschwe:
- Reproduce issue 428 by testing the legacy Generator + nightly Rust in CI.
- Refactor the CMake code for building the legacy Generator into a subdirectory
- Select a Cargo.toml / Cargo.lock for the legacy Generator depending on the Rust version

### DIFF
--- a/generator/Compat.Cargo.toml
+++ b/generator/Compat.Cargo.toml
@@ -1,0 +1,23 @@
+[package]
+name = "corrosion-generator"
+version = "0.1.0"
+authors = ["Andrew Gaspar <andrew.gaspar@outlook.com>"]
+license = "MIT"
+edition = "2018"
+
+[dependencies]
+cargo_metadata = "0.15"
+# The crates below are indirect dependencies of cargo metadata,
+# We explicitly specify maximum versions to allow building the generator
+# with older toolchains.
+# Version 1.0.157 upgrades to syn 2.0 and raises MSRV to 1.56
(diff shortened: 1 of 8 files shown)


generator/Compat.Cargo.toml
ogoffart, line 1: I don't think yo is need a Duplicated Cargo.toml
But you should have instructions on how to regenerate the Compat.Cargo.lock using the --precise flag for the crates that needs it.
What is the procedure to update the Compar.Cargo.toml in the case that, for example, the dependencies needs to be edited?

jschwe: I was considering that perhaps for newer Rust versions, we could update cargo metadata from 0.15 to 0.17 - But then I saw that cargo metadata doesn't have a changelog entry, so I decided to play it safe and not update for now.

> What is the procedure to update the Compar.Cargo.toml

I was thinking of temporarily overwriting the Cargo.toml and Cargo.lock with the `Compat` versions, running `cargo update` with whatever flags are necessary to maintain the MSRV and then copying the result back to the Compat files. 

But in practice, I think the Compat files are unlikely to ever be updated, since we have already updated as far as allowed by the MSRV restrictions (I think), and in the rust ecosystem it seems extremely unlikely to have fixes back ported to an older version of a library we are depending on. So I'm expecting them to stay as is until support for these old rust versions is removed.

PR #437, opened 2023-09-08
Fix support for hostbuild executables when cross-compiling to non-windows on windows hosts

tronical:
- Defer the determination of whether to apply a .exe suffix for the byproduct copying and imported location setting if possible
- Determine the .exe suffix first based on the existence of the hostbuild property and then fall back Rust_CARGO_TARGET_OS

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -129,6 +129,22 @@ function(_corrosion_set_imported_location_legacy target_name base_property filen
         )
 endfunction()
 
+
+# Sets out_var to true if the byproduct copying and imported location is done in a deferred
+# manner to respect target properties, etc. that may be set later.
+function(_corrosion_determine_deferred_byproduct_copying_and_import_location_handling out_var)
+    set(${out_var} ${CORROSION_RESPECT_OUTPUT_DIRECTORY} PARENT_SCOPE)
+endfunction()
+
+function(_corrosion_bin_target_suffix target_name out_var_suffix)
+    get_target_property(hostbuild "${target_name}" ${_CORR_PROP_HOST_BUILD})
+    if(hostbuild AND CMAKE_HOST_WIN32)
+        set(${out_var_suffix} ".exe" PARENT_SCOPE)
+    elseif(Rust_CARGO_TARGET_OS STREQUAL "windows")
+        set(${out_var_suffix} ".exe" PARENT_SCOPE)
+    endif()
+endfunction()
+
 # Do not call this function directly!
 #
 # This function should be called deferred to evaluate target properties late in the configure stage.
@@ -142,6 +158,16 @@ function(_corrosion_set_imported_location_deferred target_name base_property out
         set(output_dir_prop_target_name "${CMAKE_MATCH_1}")
     else()
         set(output_dir_prop_target_name "${target_name}")
+    endif()
+
+    # Append .exe suffix for executable by-products if the target is windows or if it's a host
+    # build and the host is Windows.
+    get_target_property(target_type ${target_name} TYPE)
+    if(${target_type} STREQUAL "EXECUTABLE")
+        _corrosion_bin_target_suffix(${target_name} "suffix")
+        if(suffix)
+            set(filename "${filename}${suffix}")
+        endif()
     endif()
 
     get_target_property(output_directory "${output_dir_prop_target_name}" "${output_directory_property}")
@@ -220,7 +246,8 @@ endfunction()
 #    artifact.
 # - filename of the artifact.
 function(_corrosion_set_imported_location target_name base_property output_directory_property filename)
-    if(CORROSION_RESPECT_OUTPUT_DIRECTORY)
+    _corrosion_determine_deferred_byproduct_copying_and_import_location_handling("defer")
+    if(defer)
         _corrosion_call_set_imported_location_deferred("${target_name}" "${base_property}" "${output_directory_property}" "${filename}")
     else()
         _corrosion_set_imported_location_legacy("${target_name}" "${base_property}" "${filename}")
@@ -294,6 +321,17 @@ function(_corrosion_copy_byproduct_deferred target_name output_dir_prop_name car
         endif()
     endif()
 
+    # Append .exe suffix for executable by-products if the target is windows or if it's a host
+    # build and the host is Windows.
+    get_target_property(target_type ${target_name} TYPE)
+    if(${target_type} STREQUAL "EXECUTABLE")
+        _corrosion_bin_target_suffix(${target_name} "suffix")
+        if(suffix)
+            set(tmp_file_names "${file_names}")
+            list(TRANSFORM tmp_file_names APPEND "${suffix}" OUTPUT_VARIABLE file_names)
+        endif()
+    endif()
+
     list(TRANSFORM file_names PREPEND "${cargo_build_dir}/" OUTPUT_VARIABLE src_file_names)
     list(TRANSFORM file_names PREPEND "${output_dir}/" OUTPUT_VARIABLE dst_file_names)
     message(DEBUG "Adding command to copy byproducts `${file_names}` to ${dst_file_names}")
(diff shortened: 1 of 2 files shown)


tronical:
This is an initial attempt at fixing the issue where we cross-compile on Windows to a non-Windows target but still build and call a host-build tool.

I don't quite know how to best automate testing this, but this is what I did manually:

1. Install ESP-IDF https://docs.espressif.com/projects/esp-idf/en/stable/esp32/get-started/windows-setup.html#standard-setup-of-toolchain-for-windows , a toolchain/framework for cross-compiling for Espressif MCUs.
2. Within an ESP-IDF enabled command prompt, configure Corrosion with tests enabled and `-DRust_CARGO_TARGET=tensa-esp32-none-elf`
3. Run `ctest -C Debug -R hostbuild .` to get build coverage for the described scenario. The test fails to build at the byproduct copying stage before the patch, and succeeds afterwards. The imported location path I verified via the cmake debug output Corrosion produces.

tronical:
I've now been able to verify that the imported location bit also works. The "higher level" test case is to clone Slint from git, and from an ESP-IDF prompt change into `examples/carousel/esp-idf/s3-box` and run `idf.py build`. It fails before this patch trying to copy `slint-compiler` as byproduct, when it should be `slint-compiler.exe`. After this patch the build passes.

tronical:
According to the test failures in the CI here, it seems that this requires further work.

cmake/Corrosion.cmake
```diff
@@ -321,7 +358,7 @@ function(_corrosion_copy_byproduct_deferred target_name output_dir_prop_name car
     )
 endfunction()
 
-function(_corrosion_call_copy_byproduct_deferred target_name output_dir_prop_name cargo_build_dir file_names)
+function(_corrosion_call_copy_byproduct_deferred target_name output_dir_prop_name cargo_build_dir file_names is_binary)
```
tronical, line 361: The extra parameter is rather ugly. I'm wondering if it's worth porting this over to `cmake_parse_arguments`. But I guess latest in this function it would all have to be unpacked and packed again, so maybe not worth it?

generator/src/subcommands/gen_cmake/target.rs
```diff
@@ -184,22 +184,22 @@ impl CargoTarget {
...
                 )
             endif()
             if(pdb_byproduct)
                 _corrosion_copy_byproducts(
-                    {target_name} PDB_OUTPUT_DIRECTORY \"${{cargo_build_out_dir}}\" \"${{pdb_byproduct}}\"
+                    {target_name} PDB_OUTPUT_DIRECTORY \"${{cargo_build_out_dir}}\" \"${{pdb_byproduct}}\" FALSE
                 )
             endif()
             if(bin_byproduct)
                 _corrosion_copy_byproducts(
```
tronical, line 201: I reckon this will generally go wrong, as there's no suffix handling on the caller side here. What's the status of this code path?

tronical:
Regarding testing, perhaps this is reproducible to some extent in the CI by just doing say `-DRust_CARGO_TARGET=aarch64-unknown-linux-gnu` (after installing such toolchain on Windows). I'm not sure entirely how to configure test.yaml for this.

tronical:
Yep, this is much easier to reproduce on Windows without ESP-IDF by simply adding `rustup target add aarch64-unknown-linux` and then configuring with `-DRust_CARGO_TARGET=aarch64-unknown-linux-gnu`.

tronical:
I'm going to need guidance though how to extend test.yaml best, if it's required to cover this in the CI.

generator/src/subcommands/gen_cmake/target.rs
jschwe: I was planning to remove the legacy generator for v0.5 and require a newer Cmake version (was thinking of 3.22 as in ubuntu 22.04). I could do it now.

cmake/Corrosion.cmake
jschwe: hmm, maybe we could just remove all named parameters to this function and pass [[${ARGN}]] to the deferred function.

jschwe:
Currently most of the CI logic sits in this [action.yaml](https://github.com/corrosion-rs/corrosion/blob/master/.github/actions/setup_test/action.yaml). But I would prefer to remove most of the logic from there and instead use CMake toolchain files.

Currently, the action has the assumption that host_os == target_os (and that the OS is either windows, mac or linux), so at the very least that would need to be changed. But it would probably be simpler to do the refactoring towards toolchain files instead.

>  if it's required to cover this in the CI.

I would prefer if this would be tested, otherwise it's hard to ensure that it doesn't break again.

jschwe:
Short update: I've experimented a bit with toolchain files + `CMakePresets.json` and it seems quite promising.

cmake/Corrosion.cmake
tronical: Good idea, I’ll try that.

generator/src/subcommands/gen_cmake/target.rs
tronical: I’m curious: is the newer cmake version requirement related to the removal of the generator?

jschwe: The legacy Generator enabled CMake support for CMake < 3.19 (3.19 introduced JSON parsing support). There was also some Multi-Config related code which was conditionally enabled based on CMake 3.20 / 3.21, so I decided to declare CMake 3.22 as a minimum supported version to have a bit of wiggle-room, while still supporting ubuntu-22.04 out of the box.

tronical: Thanks, makes sense to me. Looking forward to the update :)

jschwe:
Sorry for the delay, this MR fell a bit off my radard. 

> I'm going to need guidance though how to extend test.yaml best, if it's required to cover this in the CI.

Do you know what dependencies you need to install for the test and what parameters CMake needs to be configured with?
Once we have that, adding a test should be very straight forward.

tronical:
Yes, I think it’s sufficient to add `rustup target add aarch64-unknown-linux` and just configure with `-DRust_CARGO_TARGET=aarch64-unknown-linux-gnu` to reproduce the failure and see the tests pass with this PR.

jschwe:
What about installing a cross c compiler for the linking? I think the Linux targets all still default to using cc as the linker and not lld.

tronical:
I think for all tests you'd indeed need a cross c compiler for linking. But for testing this particular issue I've just used this:

1. Configure with `-DRust_CARGO_TARGET=aarch64-unknown-linux-gnu`
2. Run `ctest -C Debug -R hostbuild .`

With master it fails:

```
Test project C:/dev/corrosion/build
    Start 37: hostbuild_build
1/3 Test #37: hostbuild_build ...................***Failed   14.96 sec
    Start 38: hostbuild_run_rust-host-program
2/3 Test #38: hostbuild_run_rust-host-program ...***Not Run (Disabled)   0.00 sec
    Start 39: hostbuild_cleanup
3/3 Test #39: hostbuild_cleanup .................   Passed    0.15 sec

50% tests passed, 1 tests failed out of 2

Total Test time (real) =  15.21 sec

The following tests did not run:
         38 - hostbuild_run_rust-host-program (Disabled)

The following tests FAILED:
         37 - hostbuild_build (Failed)
Errors while running CTest
Output from these tests are in: C:/dev/corrosion/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
```

```
...
  Error copying file (if different) from "C:/dev/corrosion/build/test/hostbuild/build-hostbuild/x64/Debug/cargo/build/x86_64-pc-windows-msvc/debug/rust-host-program" to "C:/dev/corrosion/build/test/hostbuild/build-hostbuild".
...
```

But this PR it succeeds.

tronical:
Rebase pushed.

jschwe (approved):

jschwe:
Thanks!

ogoffart:
Any chances in backporting that in the 4.x branch and having a release? 
This issue is preventing our Windows users to use Slint for MCUs from C++.

PR #438, opened 2023-09-08
Require CMake 3.22

jschwe:
Remove the legacy Generator and bump the minimum required CMake version to 3.22. This allows us to remove a lot of code.

Older CMake versions will still be supported on the stable/v0.4 branch.


jschwe:
Todo: Adjust Documentation

Issue #413, opened 2023-06-25
Cross-compiling with clang and (c)dylibs in the dependency graph is broken


jschwe:
Bug Scenario: 

- Cross-compiling
- Linker driver is clang (i.e. requires --target=triple option to link for the correct target)
- OS is not MacOS (rustc adds --target automatically on `Darwin`).
- The dependency graph contains a target that requires the linker to be invoked for the cross-compilation target. This should only be dynamic libraries. Note: This does not include the final artifact - We already pass the the `--target` flag via `cargo rustc` to the final linker invocation.

Untested, but probably this requirement also applies: 
- compiler option to use lld (`-fuse-ld=lld`) is NOT set.

If such a dependency exists in the graph, then linking it will fail, since clang will call the host linker instead of the target linker.

Constraints for solutions: 

- Adding the argument to the global RUSTFLAGS has the side-effect of overriding any rustflags set via `.cargo/config.toml`, which may be unexpected, especially since it only happens in the cross-compiling with clang scenario.
- Setting the linker to `clang --target=xxx` does not work, since this is treated as the executable name.
- Writing a small wrapper shell script would be possible on Linux, but on Windows it looks like there is no guaranteed way to pass through arguments without changing them with a small script. We could instead compile a small Rust program which calls the linker-driver without going over the shell again.
- It is unclear if this will be fixed upstream in Rust, since currently gcc and clang are treated entirely the same (except on mac).

PR #444, opened 2023-09-22
Make --target link arg non-local

russelltg:
I'm running into a situation where I filter imported targets with CRATE_TYPES to _not_ include cdylib libraries, but have an executable that depends on a cdylib. Since I'm cross-compiling using clang, it needs that --target flag to compile, but since that flag is only being set locally it doesn't work. 

It's possible more flags should be set like this, but this is the one I was running into. Is this a supported usecase?

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -779,7 +779,7 @@ function(_add_cargo_build out_cargo_build_out_dir)
         # Skip adding the linker argument, if the linker is explicitly set, since the
         # explicit_linker_property will not be set when this function runs.
         # Passing this rustflag is necessary for clang.
-        corrosion_add_target_local_rustflags("${target_name}" "$<$<NOT:${explicit_linker_defined}>:${rustflag_linker_arg}>")
+        corrosion_add_target_rustflags("${target_name}" "$<$<NOT:${explicit_linker_defined}>:${rustflag_linker_arg}>")
     endif()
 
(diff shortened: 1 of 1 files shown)


jschwe:
Basically, you are hitting issue https://github.com/corrosion-rs/corrosion/issues/413, correct?

From the issue:

> Adding the argument to the global RUSTFLAGS has the side-effect of overriding any rustflags set via .cargo/config.toml, which may be unexpected, especially since it only happens in the cross-compiling with clang scenario.

Preferably I would like to solve the issue upstream, so that `rustc` automatically passes the `--target` parameter to clang when required. I already asked on zulip and there seems to be general support, but I haven't had time to work on it.

In the meantime (assuming you are on Linux), you could write a small bash script for your target platform like:

aarch64-linux-gnu-clang:
```bash
#!/usr/bin/env bash
clang  --target=aarch64-linux-gnu "${@}"
```

And then in corrosion call
```cmake
corrosion_set_linker(your_target /path/to/aarch64-linux-gnu-clang)
```

Please note that the filename of the bash script should end with `clang` (or `clang-XX`, where XX is a version number e.g. 13 or 14 ), so that  `rustc` can infer the linker type from the filename.

russelltg:
Yes, for some reason I assumed my use case was niche enough and didn't look a ton for issues, thanks!

That's a reasonable workaround. What I did it manually set that flag as non local on all the targets that it uses. (basically a workaround to implement what this PR does)

I'll close this considering it's known and you're (very reasonably) not interested in a fix within corrosion. 

Thanks!

PR #452, opened 2023-11-14
Add option to specify the Rust crate for cbindgen

Rikorose:
This is necessary when library target and crate name are different.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -1571,6 +1571,7 @@ ANCHOR: corrosion_cbindgen
 ```cmake
 corrosion_cbindgen(
         TARGET <imported_target_name>
+        CARGO_PACKAGE <cargo_package_name>
         HEADER_NAME <output_header_name>
         [MANIFEST_DIRECTORY <package_manifest_directory>]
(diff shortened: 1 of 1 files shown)


jschwe:
> Crate name is different from the library (target) name. Corrosion uses the target name while cbindgen expects the crate name.

The official documentation for crates vs packages is here: https://doc.rust-lang.org/book/ch07-01-packages-and-crates.html 
Relevant quotes: 

> A crate can come in one of two forms: a binary crate or a library crate. 

> A package is a bundle of one or more crates that provides a set of functionality. A package contains a Cargo.toml file that describes how to build those crates. 

That `cbindgen uses the `--crate` parameter and expects a package name is quite confusing. I think we should definitly stick with the official meaning of crate though.

Rikorose:
Ok, would you be fine with this PR if I change the naming convention according to the official rust/cargo documentation?
E.g. use `CARGO_PACKAGE` instead of `CRATE`?

jschwe:
Yes, that would be fine.

jschwe:
Thanks for PR! Manually squashed and merged (Not sure why github didn't detect the manual merge this time).

PR #455, opened 2023-11-14
Don't treat -framework as a library to link to, treat it as a composite arg

russelltg:
Because CMake de-duplicates libraries, we had a situation where cargo was reporting `-framework A -framework B` as its native static libs, this ended up on the link commandline as `-framework A B`, which is is not correct (the -framework flag is required). 

CMake deals with this by combining them, and they will be properly passes as separate flags on the final link commandline. 

This fixes our internal app which uses corrosion, so is at least moderately tested

Questions
* do you prefer some other control flow here?


cmake/FindRust.cmake
```diff
@@ -162,7 +162,19 @@ function(_corrosion_determine_libs_new target_triple out_libs)
...
+
+            set(was_last_framework OFF)
             foreach(lib ${libs_list})
+                # merge -framework;lib -> "-framework lib" as CMake does de-duplication of link libraries, and -framework prefix is required
+                if (lib STREQUAL "-framework")
+                    set(was_last_framework ON)
+                    continue()
+                endif()
+                if (was_last_framework)
+                    list(APPEND stripped_lib_list "-framework ${lib}")
```
jschwe, line 174:
```suggestion
                    list(APPEND stripped_lib_list "SHELL:-framework ${lib}")
```

Could you test adding the [`SHELL`](https://cmake.org/cmake/help/latest/command/add_link_options.html#option-de-duplication) prefix? I guess it is already working for you even without the `SHELL` prefix, but since the documentation says it prevents deduplication, adding it shouldn't hurt

russelltg: I was expecting that to work, but it doesn't. The cache entry ends up being

```
Rust_CARGO_TARGET_LINK_NATIVE_LIBS:INTERNAL=System;objc;SHELL:-framework Security;SHELL:-framework Foundation;c;m
```
and the link commandline ends up being `-lSHELL:-framework Security -lSHELL:-framework Foundation`

It seems like how I wrote it is the typical way to link to frameworks. I couldn't find a direct source but:
https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#id29
https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindCoin3D.cmake#L64
https://gitlab.kitware.com/cmake/cmake/-/blob/master/Tests/BundleTest/CMakeLists.txt#L44

jschwe: thanks for testing!

PR #462, opened 2023-11-28
Backport #437 windows hostbuild fix to v0.4

jschwe:
Original PR: #437

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,3 +1,9 @@
+# v0.4.5 (2ß23-11-29)
+
+### Fixes
+
+- Fix hostbuild executables when cross-compiling on windows to non-windows targets.
+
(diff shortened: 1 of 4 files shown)


jschwe:
@ogoffart @tronical Please double-check that this works as expected in slint. The patches applied cleanly and I don't believe there is any relevant difference for this fix compared to master, but I'd prefer if you checked it works before merging.

ogoffart (commented):
Thank you!

tronical:
> @ogoffart @tronical Please double-check that this works as expected in slint. The patches applied cleanly and I don't believe there is any relevant difference for this fix compared to master, but I'd prefer if you checked it works before merging.

Thank you @jschwe !! I can confirm that this back port branch fixes the issue. I did a build without it, it fails, I built with it, and now it works.

PR #472, opened 2023-12-29
Draft: Propagate rustc native-static-libs to CMake

### LINKED ISSUE #335: Parse the native staticlib dependencies output from rustc
Currently, corrosion hardcodes the libraries required by Rust static libs (OS-dependent).
This list may not be always up-to-date and cover all required libraries, since the exact list may change between different rust versions, and additional libraries may be added by `#[link]` attributes in user code or from build scripts!
Getting this information requires actually building the crate and all its dependencies, though.

Probably the only way to pass these dependencies from the rustc output at build-time to the CMake linker command line would be by saving the required arguments to a response file, which clang/gcc support loading via `@file`. I think MSVC also has such an option.

jschwe:
CMake usually requires all libraries required for linking to be
specified at CMake configure time. However, in Rust native dependencies
may be specified inside the source code via the [`link`] attribute.
`rustc` can be instructed to print all required native libraries at
build time, however this is already after the CMake configure stage
has finished. To work around this we wrap the build command for Rust
static libraries in a wrapper script, that parses the required libraries
from the `rustc` output and writes them into an `rsp` file.
At configure time we can then instruct CMake to read the linker
arguments from the file that our wrapper will generate later at build
time.

[`link`]: https://doc.rust-lang.org/reference/items/external-blocks.html#the-link-attribute

Closes #335

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -589,7 +589,7 @@ function(_add_cargo_build out_cargo_build_out_dir)
                 -P ${CORROSION_CMAKE_SRC_DIR}/CorrosionInvokeRustc.cmake
                 --)
         if(Rust_CARGO_TARGET_ENV STREQUAL "msvc")
-            target_link_libraries(${target_name} INTERFACE "@${native_libs_rsp_file}")
+            target_link_libraries(${target_name} INTERFACE "/link@${native_libs_rsp_file}")
         else()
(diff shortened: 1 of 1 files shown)


jschwe:
status: works locally on an M1 mac and x86 fedora 39, but on CI ld is behaving wierdly and not finding libraries that are definitely there. Currently don't have time to investigate.

PR #475, opened 2024-01-10
Fix _CORROSION_RUST_CARGO_TARGET_UNDERSCORE

LeonMatthesKDAB:
This variable didn't actually store an underscore value but used kebab-style, which couldn't be overwritten.


LeonMatthesKDAB:
Extracted from #474

jschwe:
thanks!

PR #477, opened 2024-01-17
Fix hostbuild when crosscompiling to windows

jschwe:
Fix hostbuild when crosscompiling to windows from a different OS.


ogoffart:
Thanks a lot @jschwe

PR #490, opened 2024-02-28
Set `_CORROSION_VERBOSE_OUTPUT_FLAG` as cache entry

mbrobbel:
The value of the `_CORROSION_VERBOSE_OUTPUT_FLAG` variable (set based on the `CORROSION_VERBOSE_OUTPUT` option) isn't propagated to the `_add_cargo_build` function where it is used: https://github.com/corrosion-rs/corrosion/blob/0a3bdf452995803c334d79dc9a2affbfad51b720/cmake/Corrosion.cmake#L768 Setting it as cache entry fixes that.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -511,10 +511,10 @@ include(CorrosionGenerator)
 # Note: `cmake_language(GET_MESSAGE_LOG_LEVEL <output_variable>)` requires CMake 3.25,
 # so we offer our own option to control verbosity of downstream commands (e.g. cargo build)
 if (CORROSION_VERBOSE_OUTPUT)
-    set(_CORROSION_VERBOSE_OUTPUT_FLAG --verbose)
+    set(_CORROSION_VERBOSE_OUTPUT_FLAG --verbose CACHE INTERNAL "")
 else()
     # We want to silence some less important commands by default.
-    set(_CORROSION_QUIET_OUTPUT_FLAG --quiet)
+    set(_CORROSION_QUIET_OUTPUT_FLAG --quiet CACHE INTERNAL "")
 endif()
 
 set(_CORROSION_CARGO_VERSION ${Rust_CARGO_VERSION} CACHE INTERNAL "cargo version used by corrosion")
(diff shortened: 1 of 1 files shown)


cmake/Corrosion.cmake
```diff
@@ -511,10 +511,10 @@ include(CorrosionGenerator)
 # Note: `cmake_language(GET_MESSAGE_LOG_LEVEL <output_variable>)` requires CMake 3.25,
 # so we offer our own option to control verbosity of downstream commands (e.g. cargo build)
 if (CORROSION_VERBOSE_OUTPUT)
-    set(_CORROSION_VERBOSE_OUTPUT_FLAG --verbose)
+    set(_CORROSION_VERBOSE_OUTPUT_FLAG --verbose CACHE INTERNAL "")
 else()
     # We want to silence some less important commands by default.
-    set(_CORROSION_QUIET_OUTPUT_FLAG --quiet)
+    set(_CORROSION_QUIET_OUTPUT_FLAG --quiet CACHE INTERNAL "")
```
jschwe, line 517:
```suggestion
    set(_CORROSION_QUIET_OUTPUT_FLAG --quiet CACHE INTERNAL "" FORCE)
```

cmake/Corrosion.cmake
```diff
@@ -511,10 +511,10 @@ include(CorrosionGenerator)
 # Note: `cmake_language(GET_MESSAGE_LOG_LEVEL <output_variable>)` requires CMake 3.25,
 # so we offer our own option to control verbosity of downstream commands (e.g. cargo build)
 if (CORROSION_VERBOSE_OUTPUT)
-    set(_CORROSION_VERBOSE_OUTPUT_FLAG --verbose)
+    set(_CORROSION_VERBOSE_OUTPUT_FLAG --verbose CACHE INTERNAL "")
```
jschwe, line 514:
```suggestion
    set(_CORROSION_VERBOSE_OUTPUT_FLAG --verbose CACHE INTERNAL "" FORCE)
```

jschwe (approved):
Thanks for the PR! I would add a `FORCE`, so that the internal cache variables are updated when `CORROSION_VERBOSE_OUTPUT` is changed.

jschwe:
Ah, true - thanks for catching that. In that case please drop the commit I added.

PR #496, opened 2024-04-02
corrosion_import_crate: Fix PROFILE argument in combination with CRATES

### LINKED ISSUE #495: corrosion_import_crate: issue with PROFILE argument in combination with CRATES
I'm trying to invoke import like so:

```cmake
corrosion_import_crate(MANIFEST_PATH ${CMAKE_CURRENT_BINARY_DIR}/git/Cargo.toml
  PROFILE release
  CRATES my-crate
)
```

however I get:

```
`corrosion_import_crate()` was called with the `CRATES` parameter set to `my-cratePROFILErelease`.
Corrosion will only attempt to import packages matching names from this list.
```

It looks like the issue is with passing `PROFILE` argument to `_generator_add_cargo_targets` which does not seem to take such argument:

https://github.com/corrosion-rs/corrosion/blob/366066c9caed822a4cbfc2fafbf7d0116d59b832/cmake/Corrosion.cmake#L930-L939

With the forwarding removed it seems to work. Is it the right thing to do?

Felix-El:
Do not forward the PROFILE argument from `corrosion_import_crate` to `_generator_add_cargo_targets`.

Closes #495.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -934,7 +934,6 @@ function(corrosion_import_crate)
             imported_crates
         ${crate_allowlist}
         ${crate_types}
-        ${cargo_profile}
         ${no_linker_override}
     )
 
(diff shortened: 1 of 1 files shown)


jschwe:
Thanks for the PR! Yes, it's a leftover from 411a7f37abf3cdc7941c70ff166a06aef19cacd3, where the `PROFILE` argument was changed to be set via target property and not propagated down the call-chain anymore.

PR #504, opened 2024-04-17
Consider vxworks executable suffix

Felix-El:
Assume .vxe suffix when target OS is vxworks


jschwe:
Looks good to me. Did you test this fix works on an actual project with vxworks?

Felix-El:
Yes sure. I wouldn't care if it wasn't for an actual failing "install" step (copy with file not found due to empty suffix)

Issue #505, opened 2024-04-22
cargo_link_libraries behavior vs. expectation


Felix-El:
I found `cargo_link_libraries` behavior, which according docs is `essentially the equivalent to target_link_libraries()`, to behave somewhat contrary to expectation.

`cargo_link_libraries` really just adds `-l <lib>` and `-L <libdir>` whereas from `target_link_libraries` I can _at least_ expect that transitive dependencies are considered (our Rust library might link some FFI).

Additionally, `target_link_libraries` also supports linking by full path and if a non-target is passed, it is assumed that it is a system library `-l <non-target-lib>`.

Should not the same be offered by `cargo_link_libraries`?

jschwe:
Sure, all of that would be great to have - pull requests are very welcome.

PR #508, opened 2024-04-23
Improvements to corrosion_link_libraries()

Felix-El:
As mentioned in #505,
- Allow linking by full path and non-target library name
- Propagate linking dependencies to artifacts

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -1077,7 +1077,6 @@ function(corrosion_link_libraries target_name)
             " aside from establishing a build dependency."
             )
     endif()
-    add_dependencies(_cargo-build_${target_name} ${ARGN})
     foreach(library ${ARGN})
         set_property(
             TARGET _cargo-build_${target_name}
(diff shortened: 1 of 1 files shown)


cmake/Corrosion.cmake
```diff
@@ -1086,8 +1085,29 @@ function(corrosion_link_libraries target_name)
...
+            corrosion_add_target_local_rustflags(${target_name}
+                "-L$<TARGET_LINKER_FILE_DIR:${library}>"
+                "-l$<TARGET_LINKER_FILE_BASE_NAME:${library}>"
+            )
+            add_dependencies(_cargo-build_${target_name} ${library})
+            get_property(libs TARGET ${library} PROPERTY INTERFACE_LINK_LIBRARIES)
+            corrosion_link_libraries(${target_name} ${libs})
+        elseif(IS_ABSOLUTE ${library})
+            # Linking via full path
+            corrosion_add_target_local_rustflags(${target_name} "-Clink-arg=${library}")
```
jschwe, line 1098: Could you add 
- a link to the relavant rustc documentatation (https://doc.rust-lang.org/rustc/command-line-arguments.html#linking-modifiers-verbatim)
- ideally a test project that utilizes absolute path linking.

cmake/Corrosion.cmake
```diff
@@ -1086,8 +1085,29 @@ function(corrosion_link_libraries target_name)
...
+        else()
+            # We have to assume ${library} is a non-CMake library name
+            corrosion_add_target_local_rustflags(${target_name} "-l${library}")
+        endif()
+
+        # Propagate dependency on ${library}
+        set_property(
+            TARGET _cargo-build_${target_name}
+            APPEND
+            PROPERTY INTERFACE_LINK_LIBRARIES
```
jschwe, line 1108: `_cargo-build_${target_name}` is a custom target, so I don't believe this property will have any effect. 
Is your intention here to add a rule to trigger rebuilds, or to propogate the dependencies to C/C++ consumers of a Rust static library ? If it's the latter, then https://github.com/corrosion-rs/corrosion/pull/506 should be sufficient.

cmake/Corrosion.cmake
```diff
@@ -1086,8 +1085,29 @@ function(corrosion_link_libraries target_name)
...
 
-        corrosion_add_target_local_rustflags(${target_name} "-L$<TARGET_LINKER_FILE_DIR:${library}>")
-        corrosion_add_target_local_rustflags(${target_name} "-l$<TARGET_LINKER_FILE_BASE_NAME:${library}>")
+        if (TARGET ${library})
+            corrosion_add_target_local_rustflags(${target_name}
+                "-L$<TARGET_LINKER_FILE_DIR:${library}>"
+                "-l$<TARGET_LINKER_FILE_BASE_NAME:${library}>"
+            )
+            add_dependencies(_cargo-build_${target_name} ${library})
+            get_property(libs TARGET ${library} PROPERTY INTERFACE_LINK_LIBRARIES)
```
jschwe, line 1094: `INTERFACE_LINK_LIBRARIES` in CMake is evaluated at build time. This means simply reading this property here would cause us to miss any libraries that are added to `INTERFACE_LINK_LIBRARIES` after `corrosion_link_libraries()`.
We could work around this by making a deferred function call for `TARGET`s, which reads the property at a later point in time.

This is just a nit, it's not required for merging this PR.

cmake/Corrosion.cmake
```diff
@@ -1086,8 +1085,29 @@ function(corrosion_link_libraries target_name)
...
+            corrosion_link_libraries(${target_name} ${libs})
```
jschwe, line 1095: Can you add a recursion check here? While ideally there would be no cyclic dependencies, I don't think that assumption holds in practice.

The implementation could probably use something like this (not tested)
 ```
set(recursion_list "${recursion_list};${library}")
if( "${library}" IN ${recursion_list})
    return
endif()
# do the things
# the pop happens implicitly, due to scoping of variable names in functions.
```

jschwe:
Partially merged via #559.  What's remaining is 

> Propagate linking dependencies to artifacts

PR #509, opened 2024-04-23
cmake export (find_package) becomes portable

thorsten-klein:
Description:
Currently the cmake export `CorrosionConfig.cmake` adds a hardcoded path (`CMAKE_INSTALL_FULL_DATADIR`) to `CMAKE_MODULE_PATH`. Unfortunately this path is only valid during build time of corrosion. 
If the resulting package is shipped (e.g. via a package manager like conan), then the path and therefore also the corrosion package becomes invalid. 

Solution:
Instead, the import prefix of the package (`PACKAGE_PREFIX_DIR`) is used.

### DIFF
--- a/cmake/CorrosionConfig.cmake.in
+++ b/cmake/CorrosionConfig.cmake.in
@@ -4,6 +4,6 @@ if (Corrosion_FOUND)
     return()
 endif()
 
-list(APPEND CMAKE_MODULE_PATH "@CMAKE_INSTALL_FULL_DATADIR@/cmake")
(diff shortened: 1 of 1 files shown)


jschwe:
Thanks, sounds reasonable. 

> e.g. via a package manager like conan

Can I assume you tested the fix and can confirm it works with conan?

thorsten-klein:
Yes, tested with cmake 3.26.3 in conan. I am able to consume corrosion now via conan (by using env variable `CMAKE_PREFIX_PATH`)

PR #511, opened 2024-04-26
msvc: handle flags that come from native-static-libs

russelltg:
As of https://github.com/rust-lang/rust/pull/122268, rust emits /defaultlib:msvcrt in native-static-libs

This ends up in INTERFACE_LINK_LIBRARIES, and then ninja thinks it's a file:

ninja: error: '/defaultlib:msvcrt', needed by 'cpp-exe.exe', missing and no known rule to make it

This check to see if libraries start with a slash, and then assuming that they are a flag, and putting those in INTERFACE_LINK_OPTIONS instead. If there is some better way to detect flags, I'm all for it. We could also have it have anything that starts with `-` but not start with `-l` be detected as a flag, but that seems a bit more likely to be disruptive.

I checked commandlines and it does indeed get propaged properly to the commandline.

### DIFF
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -177,8 +177,8 @@ function(_corrosion_determine_libs_new target_triple out_libs out_flags)
                     continue()
                 endif()
                 
-                # Flags start with /
-                if (lib MATCHES "^/")
+                # Flags start with / for MSVC
+                if (lib MATCHES "^/" AND ${target_triple} MATCHES "msvc$")
                     list(APPEND flag_list "${lib}")
                 else()
                     # Strip leading `-l` (unix) and potential .lib suffix (windows)
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -133,7 +133,7 @@ function(_corrosion_parse_target_triple target_triple out_arch out_vendor out_os
     set("${out_env}" "${CMAKE_MATCH_6}" PARENT_SCOPE)
 endfunction()
 
-function(_corrosion_determine_libs_new target_triple out_libs)
+function(_corrosion_determine_libs_new target_triple out_libs out_flags)
     set(package_dir "${CMAKE_BINARY_DIR}/corrosion/required_libs")
     # Cleanup on reconfigure to get a cleans state (in case we change something in the future)
     file(REMOVE_RECURSE "${package_dir}")
@@ -162,6 +162,7 @@ function(_corrosion_determine_libs_new target_triple out_libs)
         if(cargo_build_error_message MATCHES "native-static-libs: ([^\r\n]+)\r?\n")
             string(REPLACE " " ";" "libs_list" "${CMAKE_MATCH_1}")
             set(stripped_lib_list "")
+            set(flag_list "")
 
             set(was_last_framework OFF)
             foreach(lib ${libs_list})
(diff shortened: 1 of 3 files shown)


cmake/FindRust.cmake
```diff
@@ -175,10 +176,16 @@ function(_corrosion_determine_libs_new target_triple out_libs)
                     set(was_last_framework OFF)
                     continue()
                 endif()
-                # Strip leading `-l` (unix) and potential .lib suffix (windows)
-                string(REGEX REPLACE "^-l" "" "stripped_lib" "${lib}")
-                string(REGEX REPLACE "\.lib$" "" "stripped_lib" "${stripped_lib}")
-                list(APPEND stripped_lib_list "${stripped_lib}")
+                
+                # Flags start with /
+                if (lib MATCHES "^/")
```
jschwe, line 181: perhaps we could additionally require here that the `${target_triple}` matches `msvc$` ?

jschwe:
Could you also add an entry to the changelog in `RELEASES.md`?

jschwe:
thanks!

cmake/FindRust.cmake
```diff
@@ -762,28 +770,35 @@ set(Rust_CARGO_HOST_ENV "${rust_host_env}" CACHE INTERNAL "Host environment")
...
             "Required native libraries when linking Rust static libraries")
+    set(Rust_CARGO_TARGET_LINK_OPTIONS "${required_link_flags}" CACHE INTERNAL
+            "Required link flags when linking Rust static libraries")
 endif()
 
 if(Rust_CROSSCOMPILING AND NOT DEFINED CACHE{Rust_CARGO_HOST_TARGET_LINK_NATIVE_LIBS})
     message(STATUS "Determining required link libraries for target ${Rust_CARGO_HOST_TARGET_CACHED}")
     unset(host_libs)
-    _corrosion_determine_libs_new("${Rust_CARGO_HOST_TARGET_CACHED}" host_libs)
+    _corrosion_determine_libs_new("${Rust_CARGO_HOST_TARGET_CACHED}" host_libs host_flags)
```
wantehchang, line 794: It seems that we should add the following message after line 797, similar to lines 777-779 above:

```
    if(DEFINED host_flags)
        message(STATUS "Required link flags for host target ${Rust_CARGO_HOST_TARGET_CACHED}: ${host_flags}" )
    endif()
```

cmake/FindRust.cmake
```diff
@@ -762,28 +770,35 @@ set(Rust_CARGO_HOST_ENV "${rust_host_env}" CACHE INTERNAL "Host environment")
 if(NOT DEFINED CACHE{Rust_CARGO_TARGET_LINK_NATIVE_LIBS})
     message(STATUS "Determining required link libraries for target ${Rust_CARGO_TARGET_CACHED}")
     unset(required_native_libs)
-    _corrosion_determine_libs_new("${Rust_CARGO_TARGET_CACHED}" required_native_libs)
+    _corrosion_determine_libs_new("${Rust_CARGO_TARGET_CACHED}" required_native_libs required_link_flags)
```
wantehchang, line 773: Should we add `unset(required_link_flags)` before this line?

cmake/FindRust.cmake
wantehchang, line 794: Should we add `unset(ost_flags)` before this line?

PR #516, opened 2024-05-10
Breaking: Backport #514 to the stable branch

jschwe:
Will probably be released as v0.5, since it is a breaking change.


jschwe:
Released as stable/v0.5

PR #521, opened 2024-05-13
cbindgen: Remove CARGO_PACKAGE parameter

jschwe:
`CARGO_PACKAGE` was added in #452 to work around the package name and crate name (potentially) being different.
Since b11a33 corrosion saves the package name as a target property, which we can use in `corrosion_experimental_cbindgen`. This allows us to remove the `CARGO_PACKAGE` parameter.


jschwe:
CC @Rikorose (since the original PR was from you)

PR #537, opened 2024-06-22
Leave choosing `msvcrt` up to the C/C++ CMake code

jschwe:
It's difficult for Corrosion to correctly choose the right msvcrt version. CMake has a target property for it, and we also don't know which build configurations the user CMake code expectes to be Debug related.

The previous code actually already removed the linking against `msvcrt` in all cases except "Debug" builds. This was not intentional, but indicates that removing the library entirely from the link list, and leaving it up to the C/C++ code to link it in, should work well in practice.

### DIFF
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -190,7 +190,7 @@ function(_corrosion_determine_libs_new target_triple out_libs out_flags)
             # We leave it up to the C/C++ executable that links in the Rust static-library
             # to determine which version of the msvc runtime library it should select.
             list(FILTER libs_list EXCLUDE REGEX "^msvcrtd?")
-            list(FILTER flags_list EXCLUDE REGEX "^/defaultlib:msvcrtd?")
+            list(FILTER flag_list EXCLUDE REGEX "^/defaultlib:msvcrtd?")
         else()
             message(DEBUG "Determining required native libraries - failed: Regex match failure.")
             message(DEBUG "`native-static-libs` not found in: `${cargo_build_error_message}`")
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -187,8 +187,10 @@ function(_corrosion_determine_libs_new target_triple out_libs out_flags)
                 endif()
             endforeach()
             set(libs_list "${stripped_lib_list}")
-            # Special case `msvcrt` to link with the debug version in Debug mode.
-            list(TRANSFORM libs_list REPLACE "^msvcrt$" "\$<\$<CONFIG:Debug>:msvcrtd>")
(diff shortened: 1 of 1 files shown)


jschwe:
@Billyzou0741326 could you test this PR? I haven't tested this on a windows machine, but I believe we don't even need to specify `msvcrt` on the link line, since we should be able to rely on CMake adding `msvcrt` to the link line of the C/++ executable or shared library, that we link our Rust static library into.

cmake/FindRust.cmake
```diff
@@ -187,8 +187,10 @@ function(_corrosion_determine_libs_new target_triple out_libs out_flags)
                 endif()
             endforeach()
             set(libs_list "${stripped_lib_list}")
-            # Special case `msvcrt` to link with the debug version in Debug mode.
-            list(TRANSFORM libs_list REPLACE "^msvcrt$" "\$<\$<CONFIG:Debug>:msvcrtd>")
+            # We leave it up to the C/C++ executable that links in the Rust static-library
+            # to determine which version of the msvc runtime library it should select.
+            list(FILTER libs_list EXCLUDE REGEX "^msvcrtd?")
+            list(FILTER flags_list EXCLUDE REGEX "^/defaultlib:msvcrtd?")
```
Billyzou0741326, line 193:
```suggestion
            list(FILTER flag_list EXCLUDE REGEX "^/defaultlib:msvcrtd?")
```

A minor typo, it seems. On my windows box I'm still getting `/defaultlib:msvcrt` on the link flag due to this:

```
-- Determining required link libraries for target x86_64-pc-windows-msvc
-- Required static libs for target x86_64-pc-windows-msvc: kernel32;advapi32;kernel32;ntdll;userenv;ws2_32;kernel32;ws2_32;kernel32
-- Required link flags for target x86_64-pc-windows-msvc: /defaultlib:msvcrt
```

After fixing the typo, it filters out as expected:

```
-- Required static libs for target x86_64-pc-windows-msvc: kernel32;advapi32;kernel32;ntdll;userenv;ws2_32;kernel32;ws2_32;kernel32
-- Required link flags for target x86_64-pc-windows-msvc:
```

And then the warning goes away:

```
-- Required static libs for target x86_64-pc-windows-msvc: kernel32;advapi32;kernel32;ntdll;userenv;ws2_32;kernel32;ws2_32;kernel32
-- Required link flags for target x86_64-pc-windows-msvc:
```

Billyzou0741326:
I'm no longer seeing either `msvcrt` or `msvcrtd` being specified at all in the generated build scripts (checked both Ninja and the VS generator).

Not sure if this change will have other unintended effects, as I only tested via a trivial `add` method in the `extern "C"` style.

Billyzou0741326 (approved):

Issue #415, opened 2023-07-04
Support installing libraries


alex-semov:
### Current Behavior

When attempting to use a Rust library `my_lib` in a C++ application `my_app`, using Corrosion, the linker reports an error during the linking phase.

### Error Output:
```
[ 59%] Linking CXX executable my_app
ld: cannot find -lmy_lib-static
```

Relevant issue: https://github.com/corrosion-rs/corrosion/issues/261#issuecomment-1307223603

### Expected Behavior

The `my_app` application should successfully link with the `my_lib` Rust library without the need for renaming the output static library file.

I was able to temporarily resolve this issue by renaming the output static library as follows:

```
install(
  FILES
    ${CMAKE_CURRENT_BINARY_DIR}/libmy_lib.a
  DESTINATION
    ${lib_dest}
  RENAME
    libmy_lib-static.a
)
```

### Steps To Reproduce

Here is a snippet of the CMakeLists.txt file which reproduces the issue:

```
corrosion_import_crate(
  MANIFEST_PATH ${CARGO_MANIFEST_DIR}/Cargo.toml
  CRATES my_lib
)

target_link_libraries(my_app my_lib)

install(
  TARGETS
    my_lib
  EXPORT
    MyTargets
  DESTINATION
    ${lib_dest}/
)
```
`cmake -S. -Bbuild -DRust_CARGO_TARGET=mips-unknown-linux-gnu ...`

### Environment

```markdown
- OS: Docker, debian:buster-slim
- CMake: 3.25.0
- CMake Generator: Unix Makefiles
- Cross Compilation Toolchain: mips-unknown-linux-gnu
```

jschwe:
To clarify: the install step is essential to reproduce this issue, correct?

alex-semov:
Yes, the install step is essential to reproduce this issue. It seems that Corrosion is appending a -static suffix to the target name for static libraries, which is causing a mismatch during the linking phase because the library with the -static suffix is not being installed.

https://github.com/corrosion-rs/corrosion/blob/fc8dd409ba168c6a24572c72cf076154b913ebaa/cmake/Corrosion.cmake#L456C10-L456C10

```
if(has_staticlib)
    add_library(${target_name}-static STATIC IMPORTED GLOBAL)
    add_dependencies(${target_name}-static cargo-build_${target_name})
...
target_link_libraries(${target_name} INTERFACE ${target_name}-static)
```

jschwe:
[...]

Issue #63, opened 2020-10-07
Document corrosion_install


AndrewGaspar:
RE: #60

jschwe:
Closing. We have some documentation know.

PR #543, opened 2024-08-02
Make corrosion_install work for library files and update docs

gtker:
Works on #415 and #63.

This only installs the actual library file(s) and not extras such as headers.

Not tested on Windows for compatibility with import library files for shared libraries.

I moved the function definition out to be used as an anchor so it could be used for docs without modifying it, since I'm not entirely sure what the planned feature set of it is going to be.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -1106,11 +1106,9 @@ corrosion_install(TARGETS <target1> ... <targetN>
                   ] [...])
 ```
 * **TARGETS**: Target or targets to install.
-* **EXPORT**: NOT IMPLEMENTED.
-
-TODO: The `ARCHIVE`/`LIBRARY`/... nonsense works like `<artifact-kind> <artifact-option>...` from [`install(TARGETS)`].
-
-[`install(TARGETS)`]: https://cmake.org/cmake/help/latest/command/install.html#targets
+* **ARCHIVE**/**LIBRARY**/**RUNTIME**: Designates that the following settings only apply to that specific type of object.
+* **DESTINATION**: The subdirectory within the CMAKE_INSTALL_PREFIX that a specific object should be placed. Defaults to values from GNUInstallDirs.
+* **PERMISSIONS**: The permissions of files copied into the install prefix.
 
 ANCHOR_END: corrosion-install
 #]=======================================================================]
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -1094,9 +1094,12 @@ endfunction()
 
 #[=======================================================================[.md:
 ANCHOR: corrosion-install
+** EXPERIMENTAL **: This function is currently still considered experimental
+  and is not officially released yet. Feedback and Suggestions are welcome.
+
 ```cmake
-corrosion_install(TARGETS <target1> ... <targetN> [EXPORT <export-name>]
-                  [[ARCHIVE|LIBRARY|RUNTIME|PRIVATE_HEADER|PUBLIC_HEADER]
+corrosion_install(TARGETS <target1> ... <targetN>
+                  [[ARCHIVE|LIBRARY|RUNTIME]
                    [DESTINATION <dir>]
                    [PERMISSIONS <permissions...>]
                    [CONFIGURATIONS [Debug|Release|<other-configuration>]]
@@ -1272,10 +1275,7 @@ function(corrosion_install)
                     elseif (DEFINED COR_INSTALL_DEFAULT_PERMISSIONS)
                         set(PERMISSIONS ${COR_INSTALL_DEFAULT_PERMISSIONS})
                     else()
-                        set(
-                            PERMISSIONS
-                            ${DEFAULT_PERMISSIONS} OWNER_EXECUTE GROUP_EXECUTE WORLD_EXECUTE
-                        )
+                        set(PERMISSIONS ${DEFAULT_PERMISSIONS})
                     endif()
 
                     if (DEFINED COR_INSTALL_ARCHIVE_CONFIGURATIONS)
@@ -1287,7 +1287,7 @@ function(corrosion_install)
                     endif()
 
                     install(
-                            FILES $<TARGET_PROPERTY:is_odd-static,IMPORTED_LOCATION>
+                            FILES $<TARGET_PROPERTY:${INSTALL_TARGET}-static,IMPORTED_LOCATION>
                             PERMISSIONS ${PERMISSIONS}
                             DESTINATION ${DESTINATION}
                             ${CONFIGURATIONS}
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -1092,6 +1092,25 @@ function(corrosion_link_libraries target_name)
     endforeach()
 endfunction()
 
+#[=======================================================================[.md:
+ANCHOR: corrosion-install
+```cmake
+corrosion_install(TARGETS <target1> ... <targetN> [EXPORT <export-name>]
+                  [[ARCHIVE|LIBRARY|RUNTIME|PRIVATE_HEADER|PUBLIC_HEADER]
+                   [DESTINATION <dir>]
+                   [PERMISSIONS <permissions...>]
+                   [CONFIGURATIONS [Debug|Release|<other-configuration>]]
+                  ] [...])
+```
+* **TARGETS**: Target or targets to install.
+* **EXPORT**: NOT IMPLEMENTED.
+
+TODO: The `ARCHIVE`/`LIBRARY`/... nonsense works like `<artifact-kind> <artifact-option>...` from [`install(TARGETS)`].
+
+[`install(TARGETS)`]: https://cmake.org/cmake/help/latest/command/install.html#targets
+
+ANCHOR_END: corrosion-install
+#]=======================================================================]
 function(corrosion_install)
     # Default install dirs
     include(GNUInstallDirs)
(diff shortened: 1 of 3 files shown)


cmake/Corrosion.cmake
```diff
@@ -1092,6 +1092,25 @@ function(corrosion_link_libraries target_name)
     endforeach()
 endfunction()
 
+#[=======================================================================[.md:
+ANCHOR: corrosion-install
+```cmake
+corrosion_install(TARGETS <target1> ... <targetN> [EXPORT <export-name>]
+                  [[ARCHIVE|LIBRARY|RUNTIME|PRIVATE_HEADER|PUBLIC_HEADER]
```
jschwe, line 1099: When documenting `corrosion_install` we should take care to only document what is really working. For example, `PRIVATE_HEADER` doesn't seem to be used in the code, and also doesn't make sense, since Rust code wouldn't have private header files.
I would not document any header install options yet (and I think it would make sense to do headers in a separate PR, since some discussion is probably required)

cmake/Corrosion.cmake
```diff
@@ -1092,6 +1092,25 @@ function(corrosion_link_libraries target_name)
...
+                   [DESTINATION <dir>]
+                   [PERMISSIONS <permissions...>]
+                   [CONFIGURATIONS [Debug|Release|<other-configuration>]]
+                  ] [...])
+```
+* **TARGETS**: Target or targets to install.
```
jschwe, line 1105: I think we should clarify here that the targets must have been targets added by corrosion. Perhaps something like:

> Rust library or executable targets imported by corrosion

cmake/Corrosion.cmake
```diff
@@ -1245,6 +1257,78 @@ function(corrosion_install)
...
+                    if (DEFINED COR_INSTALL_ARCHIVE_CONFIGURATIONS)
+                        set(CONFIGURATIONS CONFIGURATIONS ${COR_INSTALL_ARCHIVE_CONFIGURATIONS})
+                    elseif (DEFINED COR_INSTALL_DEFAULT_CONFIGURATIONS)
+                        set(CONFIGURATIONS CONFIGURATIONS ${COR_INSTALL_DEFAULT_CONFIGURATIONS})
+                    else()
+                        set(CONFIGURATIONS)
+                    endif()
+
+                    install(
+                            FILES $<TARGET_PROPERTY:is_odd-static,IMPORTED_LOCATION>
```
jschwe, line 1290: it still says "is_odd" here.

doc/src/usage.md
```diff
@@ -32,6 +32,13 @@ target, see [Per Target options](#per-target-options) for details.
 [`INTERFACE`]: https://cmake.org/cmake/help/latest/command/add_library.html#interface-libraries
 [target_link_libraries]: https://cmake.org/cmake/help/latest/command/target_link_libraries.html
 
+### Install crate and headers with `corrosion_install`
+
+The default CMake [install commands] do not work correctly with the targets exported from `corrosion_import_crate()`.
```
jschwe, line 37: I think in the anchor thats imported, you define [`install(TARGETS)`] but  not [install commands].

doc/src/usage.md
```diff
@@ -32,6 +32,13 @@ target, see [Per Target options](#per-target-options) for details.
 [`INTERFACE`]: https://cmake.org/cmake/help/latest/command/add_library.html#interface-libraries
 [target_link_libraries]: https://cmake.org/cmake/help/latest/command/target_link_libraries.html
 
+### Install crate and headers with `corrosion_install`
```
jschwe, line 35: I would add an "Experimental: " here. The functionality is not complete, and we don't have tests (yet), so I would expect a few things to be broken in the beginning.

cmake/Corrosion.cmake
```diff
@@ -1245,6 +1257,78 @@ function(corrosion_install)
...
+                    endif()
+
+                    if (DEFINED COR_INSTALL_ARCHIVE_PERMISSIONS)
+                        set(PERMISSIONS ${COR_INSTALL_ARCHIVE_PERMISSIONS})
+                    elseif (DEFINED COR_INSTALL_DEFAULT_PERMISSIONS)
+                        set(PERMISSIONS ${COR_INSTALL_DEFAULT_PERMISSIONS})
+                    else()
+                        set(
+                            PERMISSIONS
+                            ${DEFAULT_PERMISSIONS} OWNER_EXECUTE GROUP_EXECUTE WORLD_EXECUTE
```
jschwe, line 1277: A static library does not need execute permissions

cmake/Corrosion.cmake
```diff
@@ -1245,6 +1257,78 @@ function(corrosion_install)
...
+                    if (DEFINED COR_INSTALL_LIBRARY_CONFIGURATIONS)
+                        set(CONFIGURATIONS CONFIGURATIONS ${COR_INSTALL_LIBRARY_CONFIGURATIONS})
+                    elseif (DEFINED COR_INSTALL_DEFAULT_CONFIGURATIONS)
+                        set(CONFIGURATIONS CONFIGURATIONS ${COR_INSTALL_DEFAULT_CONFIGURATIONS})
+                    else()
+                        set(CONFIGURATIONS)
+                    endif()
+
+                    install(
+                            IMPORTED_RUNTIME_ARTIFACTS ${INSTALL_TARGET}-shared
```
jschwe, line 1326: I know I'm the one that suggested looking into `IMPORTED_RUNTIME_ARTIFACTS`, but after reading the documentation again, I'm thinking that the usecase is meant to install an IMPORTED executable along with all the dynamic libraries it depends on.

I guess there could be Rust shared libraries, which depend on other shared libraries, so having CMake automatically also install those would be helpful. But I'm not sure if that would actually work as I'm expecting now, so I think we probably really need to test this out in an example project.

Have you tested the shared library install locally and can confirm it works in practice?

jschwe (commented):
Thanks for working on this, and especially for also adding documentation!

cmake/Corrosion.cmake
gtker: I *think*, but I'm not certain, that we'll need to accept all target kinds, since if the users want to create C++ bindings as well they will need to create a new target that depends on the rust lib target. Then in `install(TARGETS rust-lib cpp-bindings EXPORT MyProjectTargets)` we need to list all targets and their dependents, otherwise it will error.

cmake/Corrosion.cmake
gtker: Will fix.

Do you have an opinion on calling it `PUBLIC_HEADER` vs something like `PUBLIC_INCLUDES`?

It seems like [`PUBLIC_HEADER`](https://cmake.org/cmake/help/latest/prop_tgt/PUBLIC_HEADER.html) is a concept that has something to do with something Apple specific, although I have also seen people using `PUBLIC_HEADER` to install normal public header files.

cmake/Corrosion.cmake
gtker: > Have you tested the shared library install locally and can confirm it works in practice?

Yes, I have tested it on Linux.

Inside [the `is_odd` directory of my example repository](https://github.com/gtker/corrosion_export/tree/main/is_odd) I run 

```cmake
cmake -B build -S . -G Ninja && cmake --build build && cmake --install build --prefix /tmp/install
```

and it outputs

```bash
-- Install configuration: ""
-- Installing: /tmp/install/lib/libis_odd.a
-- Installing: /tmp/install/lib/libis_odd.so
```

> I guess there could be Rust shared libraries, which depend on other shared libraries, so having CMake automatically also install those would be helpful. 

I am unsure if this also installs Windows `.lib` files, but I know for sure that the manual approach doesn't, so I preferred this way.

cmake/Corrosion.cmake
gtker: Fixed.

doc/src/usage.md
gtker: Fixed. I also added the `EXPERIMENTAL` warning inside the function header found on other internal functions.

doc/src/usage.md
gtker: Fixed.

cmake/Corrosion.cmake
gtker: Fixed.

cmake/Corrosion.cmake
gtker: Fixed, but still interested in naming for headers. Is an easy fix so it could also just be discussed in the header PR.

gtker:
Not sure if I have broken the MacOS tests, but it seems unrelated.

jschwe:
The test failures on macos are unrelated to this PR.

gtker:
@jschwe Do you need anything else from me on this PR?

doc/src/usage.md
```diff
@@ -32,6 +32,15 @@ target, see [Per Target options](#per-target-options) for details.
 [`INTERFACE`]: https://cmake.org/cmake/help/latest/command/add_library.html#interface-libraries
 [target_link_libraries]: https://cmake.org/cmake/help/latest/command/target_link_libraries.html
 
+### Experimental: Install crate and headers with `corrosion_install`
+
+The default CMake [install commands] do not work correctly with the targets exported from `corrosion_import_crate()`.
+Corrosion provies `corrosion_install` to automatically install relevant files:
```
TobiasFella, line 38:
```suggestion
Corrosion provides `corrosion_install` to automatically install relevant files:
```

cmake/Corrosion.cmake
```diff
@@ -1092,6 +1092,28 @@ function(corrosion_link_libraries target_name)
...
+                  [[ARCHIVE|LIBRARY|RUNTIME]
+                   [DESTINATION <dir>]
+                   [PERMISSIONS <permissions...>]
+                   [CONFIGURATIONS [Debug|Release|<other-configuration>]]
+                  ] [...])
+```
+* **TARGETS**: Target or targets to install.
+* **EXPORT**: NOT IMPLEMENTED.
+
+TODO: The `ARCHIVE`/`LIBRARY`/... nonsense works like `<artifact-kind> <artifact-option>...` from [`install(TARGETS)`].
```
jschwe, line 1111: Could you reword this sentence? Linking to the CMake docs is fine, but the wording should be fixed before merging.

jschwe (approved):
There is still one todo in the documentation that should be addressed before merging. Otherwise, looks fine for an initial implementation. Lets see how it works in practice.
If I have some time next week, I'll try to add a test.

cmake/Corrosion.cmake
gtker: Removed TODO and reworded other parameters.

I wasn't sure how final the documentation was for unreleased functions, but I understand keeping it relatively clean.

I intend on creating a PR for installing headers soon-ish as well.

gtker:
> There is still one todo in the documentation that should be addressed before merging.

Fixed.

> If I have some time next week, I'll try to add a test.

Cool. The https://github.com/gtker/corrosion_export repo can be used with the functionality that is now done in `corrosion_install` removed.

Issue #261, opened 2022-11-08
Generated headers and CMake install


jschwe:
@trondhe @jschwe 
I have a question about the usage of this feature. Suppose I have a C++ library `my_lib` that depends on the `cxx_bridge_target`, which uses this feature to generate. Currently, the generated headers are placed under `${CMAKE_CURRENT_BINARY_DIR}/corrosion_generated` directory, and with some additional sub directory path like `cxxbridge/${cxx_target}`, which I have to read the corrosion code to figure out. 

If my C++ library `my_lib` needs to be exported and installed for external consumption, I wonder how corrosion users are expected to setup the project for `my_lib` so that the generated cxx bridge related headers can be installed and consumed by `my_lib` users. Do you have any guidance on this? Thanks.

_Originally posted by @niyue in https://github.com/corrosion-rs/corrosion/issues/244#issuecomment-1306504123_

jschwe:
It looks like this is what [`corrosion_install`](https://github.com/corrosion-rs/corrosion/blob/master/cmake/Corrosion.cmake#L1273-L1278) was meant for, however it currently only seems to support installing executable files and the rest is not implemented.

As for how you can do it in the meantime:

I would have a look here: https://cmake.org/cmake/help/latest/command/install.html#installing-files and probably also file-sets if you use at least CMake 3.23: https://cmake.org/cmake/help/latest/command/target_sources.html#file-sets

jschwe:
On the corrosion side I think it would make sense to define a header fileset on CMake >= 3.23 for the generated headers.

msvetkin:
As far as a user of the `corrision_add_cxxbridge` have access to `${cxx_target}` then it can add all necessary install rules.

The reason why I think we should not add any install rules into `corrision_add_cxxbridge` is that we don't not if user want to export this headers. The headers could be internal for the project.

@jschwe I think what we should do is set a property on `${cxx_target}` which tells where it is generated headers are placed. 

```cmake
set_target_properties(${cxx_target}
   PUBLIC
        CORROSION_HEADER_DIR ${generated_dir}/include
)
````

niyue:
Thanks so much for the prompt reply.

> It looks like this is what [corrosion_install](https://github.com/corrosion-rs/corrosion/blob/master/cmake/Corrosion.cmake#L1273-L1278) was meant for

This is what I was looking for initially, and as you said, the implementation is incomplete yet.

[...]

Issue #117, opened 2021-11-22
How to identify compiled artifacts for packaging with CPack?


apthorpe:
I am trying to build a pure Rust CLI application under CMake and use `install(TARGETS ...)` or `install(DIRECTORY ...)` to tell CPack what it needs to assemble into an installation package. I have a general CPack configuration which works with general compiled code (C, C++, Fortran, etc.) and I'm not worried about that part.

I'm using `corrosion_import_crate(MANIFEST_PATH Cargo.toml)` and it's not clear which target name I need to use with `install(TARGETS ...)` - that is, where do I find the `...`?

Example: Assume I'm building a rust application named `ostrabak`. I see Corrosion has created a build target in CMake for `cargo-build_ostrabak`. Should I be using `install(TARGETS cargo-build_ostrabak)` in this case?

I'm assuming this is more of a documentation issue than anything and that hopefully it has a simple resolution. Apologies for the dumb question; stitching together VS Code extensions, CMake, and the Rust toolchain has not been terribly straightforward and while Corrosion has helped a great deal, I'm still missing the last few pieces to make all this work. I'm more than happy to provide a working example for documentation purposes once I get this sorted out.

apthorpe:
Perhaps I should be using `corrosion_install` instead; digging through the source code I distilled out the following:

```
corrosion_install(TARGETS ostrabak
                  DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/_install
                  CONFIGURATIONS Debug)
```

but this does nothing, not even throw an error. I'm still stuck not knowing what I should use as an argument to `TARGETS`. Do I need to wrap `corrosion_import_crate()` with `add_custom_target()` to get a target name?

apthorpe:
Issue #65 seems related. Also, it doesn't appear that `-DCMAKE_BUILD_TYPE=Release` passes the build type on to `cargo`/`rustc`, defaulting to `debug` build no matter what type of build is actually specified.

My current workaround for this example is:

~~~
if(WIN32)
    install(FILES ${CMAKE_CURRENT_BINARY_DIR}/Debug/ostrabak.exe
    DESTINATION bin)
else()
    install(FILES ${CMAKE_CURRENT_BINARY_DIR}/Debug/ostrabak
    DESTINATION bin)
endif()
~~~

It's brittle and file extensions could probably be handled better but it does allow packaging to proceed which was my original issue.

jschwe:
[...]

Issue #64, opened 2020-10-07
Implement install for libraries


AndrewGaspar:
Currently `corrosion_install` only works with executables.

jschwe:
@RigoLigoRLC It's considered experimental, but corrosion_install does offer some support for installing libraries and header files. See https://corrosion-rs.github.io/corrosion/usage.html#experimental-install-crate-and-headers-with-corrosion_install

PR #544, opened 2024-08-03
Make `corrosion_install` install header files and export configurations

gtker:
This PR makes `corrosion_install` also install header files attached to the Rust target. It works both for files that are added through `target_include_directories` and through `target_sources` file sets for 3.23+.

Progresses #415, #63, #64, and #117.

I believe that this will also solve #261 since the headers are added through `target_include_directories` and `target_sources` for 3.23+, although I'm not sure how to setup cxxbridge so I can't test it.

This also correctly installs headers added through `corrosion_experimental_cbindgen`, although I had errors relating to `cbindgen` not running when testing it. If the files are generated in the correct location they will be installed.

This uses commits from #543, but since it has been approved and is just waiting for a CI run before merging I assumed it was OK to put this up.

Adding header files the old way:
```cmake
    target_include_directories(is_odd INTERFACE
            $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
            $<INSTALL_INTERFACE:include>
    )
    target_sources(is_odd INTERFACE
            $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/is_odd/is_odd.h>
            $<INSTALL_INTERFACE:include/is_odd/is_odd.h>
    )
```

will lead to

```bash
-- Installing: /tmp/install/./include
-- Installing: /tmp/install/./include/is_odd
-- Installing: /tmp/install/./include/is_odd/is_odd.h
```

while using file sets:

```cmake
    target_sources(is_odd
            INTERFACE
            FILE_SET HEADERS
            BASE_DIRS include
            FILES
            include/is_odd/is_odd.h
    )
```

will lead to
```bash
-- Installing: /tmp/install/include/is_odd/is_odd.h
```

The distinction between file sets and include directories is in preparation for `EXPORT` functionality.

# Caveat

For include directories added through `target_include_directories` the `PUBLIC_HEADER DESTINATION` argument does not determine the install location, but the name of the include directory does.
So for `target_include_directories(my-lib INTERFACE wacky-files)` and `corrosion_install(TARGETS my-lib PUBLIC_HEADER includerinos)` the file would go into `./wacky-files` rather than `./includerinos`.

This is because CMake doesn't seem to offer any real way of installing entire directories without doubling nesting them, and we can't `file(GLOB)` through the directories to find files/directories since they can be generator expressions (and must be in order to be [relocatable](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-relocatable-packages)).

From what I can tell just CMake doesn't have a good way of installing header files if you're not using 3.23+.

3.23+ file sets work completely fine.

# Open questions

* This currently installs all `PUBLIC`/`INTERFACE` file sets for a target, should it do that, or only the `HEADERS` file set?
* Is `PUBLIC_HEADER DESTINATION` not working for `target_include_directories` too confusing for it to be included? Or should we find another way. I don't have a clue how that would be done, but I'm open to suggestions.


gtker:
I'm extending this PR with functionality for `EXPORT` since it's a small change and pretty related to the header installation.

The current implementation creates a file at `${CMAKE_BINARY_DIR}/corrosion/<export-name>TargetsCorrosion.cmake` that can be included in the main config file as shown in the docs.

There's still an awful lot of boilerplate, but almost all of it is just the "normal" amount of pointless CMake configuration.

I'm creating the `COR_FILE_NAME` property on the `*-static`/`*-shared` targets in order to avoid duplicating all the logic in `corrosion_import_crate`. This will be externally visible, but I don't think there's anything wrong with leaking the file name of the library.

Sadly, because we need the base path of the install prefix in the `*Corrosion.cmake` file, users will either need to use `configure_package_config_file` or manually create the variable. I couldn't think of a better way to get it without requiring the relative destination of the file, which would scope creep the `corrosion_install` function significantly.

I have tested this with both static and shared libraries using the https://github.com/gtker/corrosion_export/ repo.

With this PR the planned functionality of `corrosion_install` should be complete.

gtker:
I completely forgot Windows IMPLIB support.

If `IMPORTED_RUNTIME_ARTIFACTS` correctly copies the IMPLIB it should work, although I can't test since I'm on Linux.

jschwe:
Haven't reviewed yet, but one short comment / question

> This is because CMake doesn't seem to offer any real way of installing entire directories without doubling nesting them

I'm not sure I understood what you mean, but did you try specifying the target directory with a trailing slash?

>  The last component of each directory name is appended to the destination directory but a trailing slash may be used to avoid this because it leaves the last component empty. 

Source: [install(DIRECTORY)](https://cmake.org/cmake/help/latest/command/install.html#directory)

cmake/Corrosion.cmake
```diff
@@ -1290,6 +1299,20 @@ function(corrosion_install)
...
+
+                    if(EXPORT_NAME)
+                        get_target_property(COR_FILE_NAME ${INSTALL_TARGET}-static COR_FILE_NAME)
+                        file(APPEND
+                            ${CMAKE_BINARY_DIR}/corrosion/${EXTRA_TARGETS_EXPORT_NAME}
+"
+add_library(${INSTALL_TARGET}-static STATIC IMPORTED)
+set_target_properties(${INSTALL_TARGET}-static
+    PROPERTIES
+    IMPORTED_LOCATION \${PACKAGE_PREFIX_DIR}/${DESTINATION}/${COR_FILE_NAME}
```
jschwe, line 1311: Anything that can be a filepath needs to be quoted, since filepaths can contain spaces.

jschwe (changes requested):
I left some comments inline

> This currently installs all PUBLIC/INTERFACE file sets for a target, should it do that, or only the HEADERS file set?

is there a need to install all file sets by default? I would start with `HEADERS` only first, unless there is a reason to install more by default.

gtker:
> I'm not sure I understood what you mean, but did you try specifying the target directory with a trailing slash?

I considered doing this like this, but we need to allow generator expressions, so it won't be a simple append-`./` operation. We could replace `>` with `./>`, but that would break the moment nested generator expressions are used.

> is there a need to install all file sets by default? I would start with HEADERS only first, unless there is a reason to install more by default.

I don't see it as completely necessary, but there had to be some default behavior so I just chose to install everything. I'll make it only install the `HEADERS` file set.

I don't have a lot of time right now, but I'll reply to your other comments later and make the changes.

I also tested it out on Windows, and I don't believe the IMPLIB is installed correctly with `IMPORTED_RUNTIME_ARTIFACTS`, so we'll need to fix that as well.

gtker:
Still working on remaining change requests, just wanted to update.

gtker:
I fixed all issues, except creating a test. 

As far as I can see there's already a test for libraries in `test/corrosion_install/`, do I need to do anything else?

jschwe:
> As far as I can see there's already a test for libraries in test/corrosion_install/, do I need to do anything else?

I probably don't have the time to review today, but perhaps you could edit the test, so that the shared library has an associated header that also needs to be installed. (https://github.com/corrosion-rs/corrosion/blob/f6ccb7105b20c62dc49cbb690e8af74ca15ee8ad/test/corrosion_install/install_lib/main.cpp#L5 this line should be in a header associated with the library, instead of hardcoded in the executable)

gtker:
I added the header file, but I really can't penetrate the meta levels of CMake script that the tests are made of. It might have been easier with python scripts that just called the cmake executable. :)

gtker:
The test failures do not appear to be mine. :)

jschwe:
Sorry for the delay, busy two weeks. I'll try to get to this MR over the weekend

jschwe (approved):
I think we can land this as is for now, and iterate over it in case any problems surface. 
Thanks a lot for looking into this.

gtker:
Cool. Let me know if any problems surface with `corrosion_install`. I won't be keeping an active eye on the issue tracker. :)

PR #553, opened 2024-08-25
Fix gnullvm and test gnullvm with msys2 in CI

jschwe:
List of problems / fixes related to gnullvm:

- [x] FindRust detects when `gnullvm` should be merged (#552).
- [x] Set correct `implib` name for `gnullvm` (#552)
- [x] Handle implib location on `gnullvm`. Ideally wait for cargo fix to be merged and reach stable. But perhaps we could also have a short term workaround if it is simple.
- [x] How to invoke the linker for Rust shared libraries?
   - [x] If corrosion uses the c/c++ linker as the linker driver like we normally do, then `-lunwind` is not found. Is the default MinGW in CI suitable?
   - [x] If we don't tell Rust which linker to use, it looks for  `x86_64-w64-mingw32-clang`, which is not in PATH, and also not the same as our C/C++ target triple. Again, raises questions on the configuration of our test case.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -143,6 +143,56 @@ jobs:
         working-directory: build
         run: ctest --output-on-failure --build-config Debug -j 3
 
+  windows_gnullvm_msys2:
+    name: Test Windows gnullvm on msys2
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - windows-2022
+        arch:
+          - x86_64
+          # - i686
+          # - aarch64
+        generator:
+          - Ninja
+          - MSYS Makefiles
+        include:
+          - arch: x86_64
+            msystem: CLANG64
+#          - arch: i686
+#            msystem: CLANG32
+#          - arch: aarch64
+#            msystem: CLANGARM64
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Rust
+        id: install_rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          targets: ${{matrix.arch}}-pc-windows-gnullvm
+      - uses: msys2/setup-msys2@v2
+        with:
+          msystem: ${{matrix.msystem}}
+          path-type: inherit
+          install: >-
+            git
+            make
+          pacboy: >-
+            toolchain:p
+            cmake:p
+            ninja:p
+      - name: Configure
+        run: cmake -S. -Bbuild -G "${{matrix.generator}}" --toolchain=.github/scripts/toolchains/${{matrix.arch}}-pc-windows-gnullvm.cmake
+      - name: Run Tests
+        working-directory: build
+        run: ctest --output-on-failure --build-config Debug -j 3
+
 # For now just test if hostbuild works when cross-compiling on windows.
 # For testing everything we would also need to install a cross-compiler first.
   windows_cross_hostbuild:
@@ -356,6 +406,7 @@ jobs:
       - visual_studio_stage2
       - windows_ninja_cl
       - windows_gnu
+      - windows_gnullvm_msys2
       - linux_stage2
       - darwin
       - test_cxxbridge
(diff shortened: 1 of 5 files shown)


mati865:
There is no LLVM toolchain preinstalled in GHA images. You can use https://github.com/mstorsjo/llvm-mingw or CLANG* environments from MSYS2: https://github.com/msys2/setup-msys2

PieroV:
>     * [ ]  How to invoke the linker for Rust shared libraries?
>       
>       * [ ]  If corrosion uses the c/c++ linker as the linker driver like we normally do, then `-lunwind` is not found. Is the default MinGW in CI suitable?
>       * [ ]  If we don't tell Rust which linker to use, it looks for  `x86_64-w64-mingw32-clang`, which is not in PATH, and also not the same as our C/C++ target triple. Again, raises questions on the configuration of our test case.

In my tests I added the toolchain to `$PATH`, but I guess the right thing would be having Cmake telling Cargo/Rust to use that?
Apart from that, that linker should be correct.

Anyway, I can any needed tests.
However, it'll take me a bit longer if I need to build a patched cargo :sweat_smile:.

jschwe:
> There is no LLVM toolchain preinstalled in GHA images

When looking at the [action logs](https://github.com/corrosion-rs/corrosion/actions/runs/10547536528/job/29220354307?pr=553) it seems that CMake does detect clang installed under `C:/Program Files/LLVM/bin/clang.exe`

```
-- The C compiler identification is Clang 18.1.8
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/LLVM/bin/clang.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
```

Is there any difference between that LLVM and what `llvm-mingw` / msys2-llvm provide?
The [official runner software list](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md) lists `Microsoft.VisualStudio.Component.VC.Llvm.Clang`. So I'm wondering what the difference between microsoft LLVM and other LLVM toolchains is (and if I can detect that from CMake to decide on the toolchain)

PieroV:
> Is there any difference between that LLVM and what `llvm-mingw` / msys2-llvm provide? The [official runner software list](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md) lists `Microsoft.VisualStudio.Component.VC.Llvm.Clang`. So I'm wondering what the difference between microsoft LLVM and other LLVM toolchains is (and if I can detect that from CMake to decide on the toolchain)

As far as I know, the binaries provided by LLVM include only the compilers and other tools (linker, windres, etc...), they don't include any runtime and/or standard library implementation.
They will look for an existing Visual Studio installation by default on Windows, but it should work also with mingw or even be able to cross compile with the proper incantation (a mix of `--sysroot`, `--target`, and similar arguments).

The llvm-mingw binaries provide an LLVM toolchain plus mingw, libc++ (LLVM's C++ standard library implementation), libunwind, compiler-rt, and maybne some other runtime.
Theoretically, you could use the official binaries with llvm-mingw's runtimes, but I don't know if there's a reason to prefer it over using only llvm-mingw for this use-case.

mati865:
> should work also with mingw or even be able to cross compile with the proper incantation (a mix of `--sysroot`, `--target`, and similar arguments).

Some features will be missing when using mingw-w64 built with GCC though, like Control Flow Guard support. Also, libraries like winpthreads won't be fully compatible because of different TLS and unwinders. This should be fine for most users, but it's better to use purposely built mingw-w64 if doable.

.github/workflows/test.yaml
```diff
@@ -36,18 +36,22 @@ jobs:
...
+        id: install_rust
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          targets: ${{matrix.arch}}-pc-windows-gnullvm
       - uses: msys2/setup-msys2@v2
         with:
-          # Note: perhaps needs to be part of the matrix if we cross-compile
-          msystem: CLANG64
+          msystem: ${{matrix.msystem}}
```
mati865, line 47:
```suggestion
          msystem: ${{matrix.msystem}}
          path-type: inherit
```

PR #555, opened 2024-09-02
corrosion_install: fix reconfiguring breaking EXPORTs

gtker:
Sorry for the immediate do over, but it seems there were a few mistakes in my last PR #544.

It seems the else was accidentally put on the check for a `TARGET ${TARGET_NAME}-static` lib instead of the `TARGET_TYPE STREQUAL *`.

Additionally, I forgot to put a guard on the `file(APPEND)` statements, so every reconfigure would add yet another set of `add_library`s for both static/shared, leading to it not working.

I'm not sure what the correct solution for this is, but the current solution at the very least works.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -1101,7 +1101,7 @@ ANCHOR: corrosion-install
   and is not officially released yet. Feedback and Suggestions are welcome.
 
 ```cmake
-corrosion_install(TARGETS <target1> ... <targetN> [EXPORT <export-name]
+corrosion_install(TARGETS <target1> ... <targetN> [EXPORT <export-name>]
                   [[ARCHIVE|LIBRARY|RUNTIME|PUBLIC_HEADER]
                    [DESTINATION <dir>]
                    [PERMISSIONS <permissions...>]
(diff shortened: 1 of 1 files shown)


cmake/Corrosion.cmake
```diff
@@ -1308,19 +1308,21 @@ function(corrosion_install)
 
                     if(EXPORT_NAME)
                         get_target_property(COR_FILE_NAME ${INSTALL_TARGET}-static COR_FILE_NAME)
-                        file(APPEND
-                            ${CMAKE_BINARY_DIR}/corrosion/${EXTRA_TARGETS_EXPORT_NAME}
+                        if(NOT EXISTS ${CMAKE_CURRENT_BINARY_DIR}/corrosion/${EXTRA_TARGETS_EXPORT_NAME}-static-exists)
```
jschwe, line 1311: Can we assume that `corrosion_install()` will only be called once per target? If yes perhaps we could delete the file first.
We would still want to overwrite the file in case `$DESTINATION`  changed.

gtker: I guess we could, but the file is at the EXPORT level and we don't know if other installs put something else in there.

I'm not entirely sure what the completely correct solution is here.

jschwe: I pushed a suggestion - If we require that the export name is unique, then we shouldn't have any problems with other `corrosion_installs` putting information into the same export.
What do you think?

gtker: I think that makes it good enough. The worst that happens is that people need to add another line to their config file.

PR #567, opened 2024-10-06
Propagate Apple deployment target version from CMake to `cc`

IgKh:
`rustc` always targets the earliest deployment platform version possible for Rust code, but the `cc` crate defaults to the host's SDK version which is typically very recent.

When the CMake uses a toolchain that has its' own minimal supported macOS version (for example Qt), this leads to linker errors and possible crashes when deploying to older yet supported systems.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -703,6 +703,11 @@ function(_add_cargo_build out_cargo_build_out_dir)
         list(APPEND corrosion_cc_rs_flags "SDKROOT=${CMAKE_OSX_SYSROOT}")
     endif()
 
+    # Ensure that cc-rs targets same Apple platform version as the CMake build
+    if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Darwin" AND CMAKE_OSX_DEPLOYMENT_TARGET)
+        list(APPEND corrosion_cc_rs_flags "MACOSX_DEPLOYMENT_TARGET=${CMAKE_OSX_DEPLOYMENT_TARGET}")
+    endif()
+
     corrosion_add_target_local_rustflags("${target_name}" "$<$<BOOL:${corrosion_link_args}>:-Clink-args=${corrosion_link_args}>")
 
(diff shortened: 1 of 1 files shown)


IgKh:
Linker warnings look like:

```
ld: warning: object file (libtypstdriver_rs.a[118](246b3582b3de8cce-cxx.o)) was built for newer 'macOS' version (14.5) than being linked (14.0)
ld: warning: object file (libtypstdriver_rs.a[206](e4479bd7784abb10-aarch_aapcs64.o)) was built for newer 'macOS' version (14.5) than being linked (14.0)
```

cmake/Corrosion.cmake
```diff
@@ -703,6 +703,11 @@ function(_add_cargo_build out_cargo_build_out_dir)
         list(APPEND corrosion_cc_rs_flags "SDKROOT=${CMAKE_OSX_SYSROOT}")
     endif()
 
+    # Ensure that cc-rs targets same Apple platform version as the CMake build
+    if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Darwin" AND CMAKE_OSX_DEPLOYMENT_TARGET)
```
jschwe, line 707:
```suggestion
    if(CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND CMAKE_OSX_DEPLOYMENT_TARGET)
```

It should only matter that CMake is targetting a MacOS system, and not if the host system is macos, right?

IgKh: I think you are right. Despite 'OSX' in the name CMake uses the `CMAKE_OSX_DEPLOYMENT_TARGET` variable for all Apple targets (also iOS, etc), but `cc-rs` is [sensitive](https://github.com/rust-lang/cc-rs/blob/218365e5837815124b4b30c0ebb8149e1cec092a/src/lib.rs#L3996) to the `IPHONEOS_DEPLOYMENT_TARGET` only if the target is actual macOS.

jschwe:
Thanks!

PR #574, opened 2024-10-30
Fix corrosion_experimental_cbindgen with a non-existent target.

### LINKED ISSUE #562: Cannot use corrosion_experimental_cbindgen with a rlib target
I am converting code from C to rust, placing the converted code into a rlib crate (foo-c).

cmake needs to produce 2 staticlibs of the converted code, for stable and unstable versions of the code.
For that I have 2 staticlib crates that just propagate features and export public symbols (foo-c-stable and foo-c-unstable)

The crates are part of a workspace and imported at the same time with a single `corrosion_import_crate` call:
 * foo-c (rlib)
 * foo-c-stable (staticlib)
 * foo-c-unstable (staticlib)
 * bin-that-only-needs-the-converted-code (bin)

The C code needs to use the generated header of the converted code.
`corrosion_experimental_cbindgen` is used on foo-c but errors in the `get_target_property` line of:
https://github.com/corrosion-rs/corrosion/blob/3db4ffdbe0e54de934f91a981b0e1d0da538408e/cmake/Corrosion.cmake#L1826-L1836
TARGET is required and MANIFEST_DIRECTORY is provided... so I guess the `else()` branch is supposed to create a target?

After the problem is fixed I need to figure out the proper way to make the C code use the generated header, probably by modifying the foo-c-stable and foo-c-unstable targets in some way. Some guidance is welcome. :)

#159 seems related to this

---

**My current workaround:**

The foo-c crate is also a staticlib for the stable code and is used to generate the header.
The foo-c-unstable target has the expected header dir manually added with `target_include_directories(INTERFACE)`.
The foo-c-unstable target manually depends on the internal "bindings" target of the foo-c target. (hack)

I want to use the workspace organization that I described, not this workaround,

flaviojs:
This assumes that corrosion_experimental_cbindgen is supposed to work with a non-existent target.
Includes test.

Fixes #562

Maybe the package name should be an argument to avoid having to parse the metadata? Design decisions are not for me.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -1836,6 +1836,28 @@ function(corrosion_experimental_cbindgen)
         else()
             set(package_manifest_dir "${CCN_MANIFEST_DIRECTORY}")
         endif()
+
+        # create non-existent target
+        add_library("${rust_target}" INTERFACE)
+
+        # get package name
+        file(REAL_PATH "${CCN_MANIFEST_DIRECTORY}/Cargo.toml" manifest_path)
+        _cargo_metadata(json "${manifest_path}")
+
+        string(JSON packages GET "${json}" "packages")
+        string(JSON pkgs_len LENGTH "${packages}")
+        math(EXPR pkgs_len-1 "${pkgs_len} - 1")
+
+        foreach(ix RANGE ${pkgs_len-1})
+            string(JSON pkg GET "${packages}" ${ix})
+            string(JSON pkg_name GET "${pkg}" "name")
+            string(JSON pkg_manifest_path GET "${pkg}" "manifest_path")
+            if(pkg_manifest_path STREQUAL manifest_path)
+                # found the package
+                set_target_properties("${rust_target}" PROPERTIES COR_CARGO_PACKAGE_NAME "${pkg_name}")
+                break()
+            endif()
+        endforeach()
     endif()
 
     get_target_property(rust_cargo_package "${rust_target}" COR_CARGO_PACKAGE_NAME )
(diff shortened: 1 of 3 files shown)


jschwe (changes requested):
I'm not a fan of adding a dummy library target. 
As I've mentioned before the code in this function is not particularly tied to corrosion - One option I could see working out is making the `TARGET` parameter of this function optional. If `TARGET` is not passed (e.g. because there is no such imported target), then the required information (mainly the package name) could be passed in by the user as optional parameters. 
Just search for all occurences of `get_target_property()` and `$<TARGET_PROPERTY:...>` and allow the respective values to be set by the user via an input parameter. 

In the case of lines 1821-1822 I would suggest to replace it with 
```
if(DEFINED CCN_TARGET)
    set(hostbuild_override "$<BOOL:$<TARGET_PROPERTY:${rust_target},${_CORR_PROP_HOST_BUILD}>>")
    set(cbindgen_target_triple "$<IF:${hostbuild_override},${_CORROSION_RUST_CARGO_HOST_TARGET},
else()
    set(cbindgen_target_triple "${_CORROSION_RUST_CARGO_TARGET}")
endif)
````

flaviojs:
Erm, the dummy target is the way other code has access to the header.
If there is no target, how should the header become available to c/c++?

The include dir is an internal detail, but it has to be exposed somehow to code that needs to use the library.
To my understanding, the cmake way of doing things is avoiding variables and using dummy targets that fully integrate with the build system.
How is MANIFEST_DIRECTORY actually supposed to work?

---

Although you say the code in this function is not particularly tied to corrosion...
A rust library without the header that describes the symbols it provides is kinda useless.
If one of the aims of corrosion is to allow easy use of rust libraries, then handling header generation seems like a requirement to me.

cxxbridge produces a header that is tightly linked to the compiled library.
It makes sense to only have it available in the target of the compiled library.

cbindgen is different, it does not need the library to be compiled (parses the source) and produces a header that works with all possible library compilations (features in rust).
It makes sense to have it available as a separate target or have it linked to all targets that represent the library.

jschwe:
Closing in favor of https://github.com/corrosion-rs/corrosion/pull/596, which adds a second signature to `corrosion_experimental_cbinden`, where all necessary parameters can be passed in by the caller - including the interface library target the generated headers should be attached to.
See https://github.com/corrosion-rs/corrosion/blob/master/test/cbindgen/manual/CMakeLists.txt for an example usage.

PR #577, opened 2024-11-13
corrosion_add_cxxbridge: Add target to regenerate bridge files

LeonMatthesKDAB:
I'm currently using CXX with Corrosion on a project that is rather difficult to build outside of a docker container and also takes a long time to build.

To get good IDE completion with clangd however, clangd needs to be able to access the header files that are generated by CXX.

This PR allows users to specify a custom CMake target that will regenerate the cxx bridges without needing to recompile.
That way, I can get good IDE completion without having to have a fully working build setup in and outside my docker container.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -1523,6 +1523,7 @@ ANCHOR: corrosion_add_cxxbridge
 ```cmake
 corrosion_add_cxxbridge(cxx_target
         CRATE <imported_target_name>
+        REGEN_TARGET <regen_target_name>
         [FILES <file1.rs> <file2.rs>]
 )
 ```
@@ -1533,6 +1534,7 @@ Adds build-rules to create C++ bindings using the [cxx] crate.
 * `cxxtarget`: Name of the C++ library target for the bindings, which corrosion will create.
 * **FILES**: Input Rust source file containing #[cxx::bridge].
 * **CRATE**: Name of an imported Rust target. Note: Parameter may be renamed before release
+* **REGEN_TARGET**: Name of a custom target that will regenerate the cxx bindings **without** recompiling. Note: Parameter may be renamed before release
 
 #### Currently missing arguments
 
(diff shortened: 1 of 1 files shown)


cmake/Corrosion.cmake
```diff
@@ -1690,6 +1692,9 @@ function(corrosion_add_cxxbridge cxx_target)
             $<INSTALL_INTERFACE:include>
     )
 
+    if(DEFINED _arg_REGEN_TARGET)
```
jschwe, line 1695: I guess this empty if was accidentally committed?

cmake/Corrosion.cmake
```diff
@@ -1739,16 +1746,22 @@ function(corrosion_add_cxxbridge cxx_target)
...
 
-        target_sources(${cxx_target}
-            PRIVATE
-                "${header_placement_dir}/${cxx_header}"
-                "${generated_dir}/include/rust/cxx.h"
-                "${source_placement_dir}/${cxx_source}"
-        )
+        list(APPEND GENERATED_FILES
+            "${header_placement_dir}/${cxx_header}"
+            "${generated_dir}/include/rust/cxx.h"
```
jschwe, line 1754: I think this line can be removed, since cxx.h is already added in the  set command a couple of lines above.

jschwe:
Seems reasonable, thanks for the PR

cmake/Corrosion.cmake
LeonMatthesKDAB: Ah, yes, will fix.

cmake/Corrosion.cmake
LeonMatthesKDAB: Hm, seems like you're right.
The previous code included it there for some reason, but I think now it shouldn't be necessary anymore :thinking: 
Will fix.

LeonMatthesKDAB: Done

LeonMatthesKDAB:
@jschwe Thank you for the review. The issues you noticed should now be fixed :)

PR #586, opened 2024-12-18
Docs: "Cannot find source file" in common_issues

LeonMatthesKDAB:
As described in #580, this error may be encountered when using corrosion_add_cxxbridge, now that the generated headers are added as `PUBLIC` sources.


jschwe (approved):
Thanks!

PR #598, opened 2025-01-01
Add crate features to check cxx version command

### LINKED ISSUE #597: `corrosion_add_cxxbridge` does not respect crate features
```cmake
corrosion_import_crate(
        MANIFEST_PATH example_path/Cargo.toml
        FEATURES interop) # Note this feature flag here
        
corrosion_add_cxxbridge(example-bridge
        CRATE example
        MANIFEST_PATH example_path
        FILES ffi.rs)
```

```toml
[package]
name = "example"
version = "0.1.0"
edition = "2021"

[lib]
crate-type = ["cdylib", "lib"]

[features]
interop = ["dep:cxx", "dep:cxx-build"] # We lock cxx features behind this feature flag.

[dependencies]
cxx = { version = "1.0.135", optional = true }

[build-dependencies]
cxx-build = { version = "1.0.135", optional = true }
```

https://github.com/corrosion-rs/corrosion/blob/999f93b98e965dc418b43854ad321a4e45385983/cmake/Corrosion.cmake#L1674C1-L1680C6

The command that is run in the code linked above (`/Users/.../.rustup/toolchains/nightly-aarch64-apple-darwin/bin/cargo tree -i cxx --depth=0`) returns the error described below, because cxx is locked behind the `interop` feature flag.

```
error: package ID specification 'cxx' did not match any packages

	Did you mean 'syn'?
```

It can be fixed by adding the features that are passed to `corrosion_import_crate` to the `cargo tree` command, so it looks like this: `/Users/.../.rustup/toolchains/nightly-aarch64-apple-darwin/bin/cargo tree -i cxx --features interop --depth=0` which then returns the correct cxx version: `cxx v1.0.135`

## Why lock cxx behind a feature flag?
My crate is being cross-compiled to many different platforms and contains lots of shared code for either WebAssembly (wasm-bindgen) and C++ (cxx) which needs to be toggled based on where it's built.

FabianHummel:
Adds the supplied crate features to the cxx version check command so it does not fail when the cxx dependency is made optional and locked behind a feature flag.

Fixes #597

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -1671,9 +1671,18 @@ function(corrosion_add_cxxbridge cxx_target)
 
     get_filename_component(manifest_dir ${manifest_path} DIRECTORY)
 
+    get_target_property(features "${_arg_CRATE}" CORROSION_FEATURES)
+    
+    # If features are empty, no need to pass them to cargo
+    if(features)
+        set(features "--features=${features}")
+    else()
+        set(features "")
+    endif()
+    
     execute_process(COMMAND ${CMAKE_COMMAND} -E env
         "CARGO_BUILD_RUSTC=${_CORROSION_RUSTC}"
-        ${_CORROSION_CARGO} tree -i cxx --depth=0
+        ${_CORROSION_CARGO} tree -i cxx ${features} --depth=0
         WORKING_DIRECTORY "${manifest_dir}"
         RESULT_VARIABLE cxx_version_result
         OUTPUT_VARIABLE cxx_version_output


jschwe:
Is this fix sufficient? In particular are the generated bindings correct, even without us passing any features to the actual `cxxbridge` command? `cxxbridge` just has a `--cfg` flag and not a feature flag, so I'm wondering if it internally uses something like `--all-features`, or if we would need to manually set the corresponding feature flags via `--cfg`. 

If `cxxbridge` uses `--all-features`, then we might also want to just use `--all-features` here too (which should be fine as long as features are really additive)

FabianHummel:
I noticed that, too. The cxxbridge command would also need to be able to accept a list of features, but I have not really looked into this yet as this is not corrosion specific.

The goal would be to achieve being able to add a conditional flag above `cxx::bridge` like in the example below, which is not yet possible with the changes I have made and requires an additional PR on cxx's side:

```rust
#[cfg(feature = "interop")]
#[cxx::bridge]
mod ffi {
   // ...
}
```

jschwe:
> The cxxbridge command would also need to be able to accept a list of features, but I have not really looked into this yet as this is not corrosion specific.

In theory, the `--cfg` option is all that is needed - it's just not super convenient - especially if we want to support `--all-features`

jschwe (changes requested):
Upon second thought, I realized we can't merge this as is. 
The reason is that Corrosion explicitly allows the features to contain CMake Generator expressions. Since `execute_process` runs at configure time, we can't use generator expressions here (at least not without more changes). 

As a quick fix, I think we could just pass `--all-features` unconditionally. 
What do you think?

FabianHummel:
Yes, passing `--all-features` is fine.

However, how does the `--cfg` flag even work? It expects `<name="value" | name[=true] | name=false>` but I have not gotten anything to work after trying several options such as `--cfg features="interop"`, `--cfg --features="interop"` or `--cfg all-features=true`

jschwe:
> However, how does the --cfg flag even work? 

it basically maps to `#[cfg(feature="xxx-feature")]`, so it needs to be singular, and I think the quotes are important too (but its been a while since I last used `--cfg`. There is also no such thing as `--cfg all-features`, we would need to parse the output of `cargo metadata` to get a list of all available features.

jschwe:
> Yes, passing --all-features is fine.

This was added /fixed in #607. My apologies for the delay - other things came up, and I somewhat forgot about your issue / PR

FabianHummel:
Alright, thanks :)

PR #601, opened 2025-01-09
Add cbindgen target to ALL

Bktero:
Hello

I ran into an issue while installing a library with a header file generated with cbindgen.

I have a minimalist project to reproduce the problem here: https://github.com/Bktero/corrosion-rs-example

It is currently using the branch of my fork I am proposing in this PR, and it works fine with it. However, if I change the `CMakeLists.txt` to use https://github.com/corrosion-rs/corrosion.git with the master branch (the latest commit being 3985d57e03acb99f1e6089fa51d072c285827b29 as I write this message), I get an error:

```shell
CMake Error at cmake_install.cmake:55 (file):
file INSTALL cannot find
"/home/pierre/bakasables/bakasable-rs/cmake-build/corrosion_generated/cbindgen/bakasable_rs/include//bakasable.h":
No such file or directory.
```

It seems that the `install` target will execute only targets that are parts of ALL. As a consequence, `make install` doesn't trigger the generation of the header. In this PR, I simply added the cbindgen target to ALL, but I don't know whether this is the most accurate solution.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -2049,7 +2049,7 @@ function(corrosion_experimental_cbindgen)
 
     set(corrosion_generated_dir "${CMAKE_CURRENT_BINARY_DIR}/corrosion_generated")
     set(generated_dir "${corrosion_generated_dir}/cbindgen/${cbindgen_bindings_target}")
-    set(header_placement_dir "${generated_dir}/include/")
+    set(header_placement_dir "${generated_dir}/include")
     set(depfile_placement_dir "${generated_dir}/depfile")
     set(generated_depfile "${depfile_placement_dir}/${output_header_name}.d")
     set(generated_header "${header_placement_dir}/${output_header_name}")
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -2113,6 +2113,7 @@ function(corrosion_experimental_cbindgen)
     # Users might want to call cbindgen multiple times, e.g. to generate separate C++ and C header files.
     string(MAKE_C_IDENTIFIER "${output_header_name}" header_identifier )
     add_custom_target("_corrosion_cbindgen_${cbindgen_bindings_target}_bindings_${header_identifier}"
+            ALL
             DEPENDS "${generated_header}"
             COMMENT "Generate ${generated_header} for ${cbindgen_bindings_target}"
     )


jschwe:
With which cmake version did the original problem occur? Was it older than cmake 3.23?

Bktero:
```
 % cmake --version
cmake version 3.25.1

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

I'm using Debian Bookworm.

jschwe:
>  In this PR, I simply added the cbindgen target to ALL, but I don't know whether this is the most accurate solution.

I'll try to reproduce this issue on the weekend and figure out what is going wrong. I'm not sure if simply adding the target to ALL can complelty solve this problem, or if it would re-appear when using `-j`.

Bktero:
It seems to work fine on my small project:

```shell
$ make install -j
[  0%] Built target cargo-prebuild_bakasable_rs
[ 50%] Generate /home/pierre/bakasables/bakasable-rs/cmake-build/corrosion_generated/cbindgen/bakasable_rs/include/bakasable.h for bakasable_rs
[100%] Built target _corrosion_cbindgen_bakasable_rs_bindings_bakasable_h
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.05s
Copying byproducts `libbakasable_rs.a` to /home/pierre/bakasables/bakasable-rs/cmake-build
[100%] Built target _cargo-build_bakasable_rs
[100%] Built target cargo-build_bakasable_rs
Install the project...
-- Install configuration: ""
-- Up-to-date: /home/pierre/bakasables/bakasable-rs/install/lib/libbakasable_rs.a
-- Up-to-date: /home/pierre/bakasables/bakasable-rs/install/include/bakasable.h
```

Let me know if I can help.

jschwe:
Your approach works fine for `make install`, because the main install target depends on ALL. 
However, if you just install one specific target, then ALL would not be required, so you would run into the same project again.

I opened https://github.com/corrosion-rs/corrosion/pull/604, which should ensure that the header is always built alongside the rust library target. This seems to work well locally (but it seems I'll need to iterate a bit over the test I added before it can be merged)

jschwe:
@Bktero I merged your "Remove extra separator in path" commit and also #604, so the issue should be resolved now. 
Please let me know if you still encounter any issues.

Bktero:
I will test it and I will let you know! Thanks.

PR #609, opened 2025-03-12
Build different packages in different folders

Algunenano:
Makes `cargo_target_dir` dependent on the package name.

If you build multiple packages using Corrosion inside the same project the would use they end up using the same target directory. Although this is apparently safe, it mixes their dependencies and confuses cargo into thinking that the dependencies have changed (because the files modification time has changed) and it needs to rebuild the packages.

I'm not sure if it makes sense for this to be optional or not. Any thoughts?

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -734,7 +734,7 @@ function(_add_cargo_build out_cargo_build_out_dir)
     set(default_build_type_dir "$<IF:$<OR:$<CONFIG:Debug>,$<CONFIG:>>,debug,release>")
     set(build_type_dir "$<IF:${cargo_profile_set},${custom_profile_build_type_dir},${default_build_type_dir}>")
 
-    set(cargo_target_dir "${CMAKE_BINARY_DIR}/${build_dir}/cargo/build")
+    set(cargo_target_dir "${CMAKE_BINARY_DIR}/${build_dir}/cargo/build/${package_name}")
     set(cargo_build_dir "${cargo_target_dir}/${target_artifact_dir}/${build_type_dir}")
     set("${out_cargo_build_out_dir}" "${cargo_build_dir}" PARENT_SCOPE)
 


jschwe:
> If you build multiple packages using Corrosion inside the same project the would use they end up using the same target directory.

At the very least for workspaces this is desired, since this matches the behavior of cargo and not doing it would cause an explosion of the required diskspace and cold compile times in larger workspaces. 

>Although this is apparently safe, it mixes their dependencies and confuses cargo into thinking that the dependencies have changed (because the files modification time has changed) and it needs to rebuild the packages.

To clarify - are you importing multiple packages that are not in one workspace?

Algunenano:
> To clarify - are you importing multiple packages that are not in one workspace?

Yes. Technically I'm importing a workspace with 2 packages on one side, and a single standalone package separately (which I can't add to the workspace and needs to be build separately).

If I only build one of them (the workspace or the standalone package) it's fine, but when both are built they modify the same files and cargo believes the project needs to be rebuilt all the time.

> At the very least for workspaces this is desired, since this matches the behavior of cargo and not doing it would cause an explosion of the required diskspace and cold compile times in larger workspaces.

This makes sense as the workspace has a common toml/lock file so it won't generate conflicts.

Maybe instead of doing this it should be configurable so it's possible to set `cargo_target_dir` to something like `${CMAKE_BINARY_DIR}/${build_dir}/cargo/build/workspace_1` for all the projects under the workspace, and to `${CMAKE_BINARY_DIR}/${build_dir}/cargo/build/workspace_2` for the second one. That way there are no conflicts between them. Would that make sense?

Algunenano:
An alternative that works in both cases (at least my 2 cases) is to use a path based on the manifest path. That way packages sharing the same path (in the same workspace) will be built in the same folder, but it will use a different one for other packages:

```patch
$ git diff
diff --git a/cmake/Corrosion.cmake b/cmake/Corrosion.cmake
index 34a8fa9..976f2b0 100644
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -734,7 +734,8 @@ function(_add_cargo_build out_cargo_build_out_dir)
     set(default_build_type_dir "$<IF:$<OR:$<CONFIG:Debug>,$<CONFIG:>>,debug,release>")
     set(build_type_dir "$<IF:${cargo_profile_set},${custom_profile_build_type_dir},${default_build_type_dir}>")
 
-    set(cargo_target_dir "${CMAKE_BINARY_DIR}/${build_dir}/cargo/build")
+    string(SHA1 CARGO_TARGET_SUBDIR ${workspace_manifest_path})
+    set(cargo_target_dir "${CMAKE_BINARY_DIR}/${build_dir}/cargo/build/${CARGO_TARGET_SUBDIR}")
     set(cargo_build_dir "${cargo_target_dir}/${target_artifact_dir}/${build_type_dir}")
     set("${out_cargo_build_out_dir}" "${cargo_build_dir}" PARENT_SCOPE)
```

Would this be preferable?

jschwe:
Sorry for the delay. I would prefer for the path to be human-readable - not a great fan of adding a hash to the path. As a user, I would prefer to be able to tell at a glance which build folder is for what (and a hash doesn't tell me anything). 

I guess the issue we need to solve is that there is no such thing as a unique "workspace name". Taking the folder name of the workspace manifest should likely work out in practice, though. We could add a parameter allowing the user to override if the folder name does occur multiple times.

Algunenano:
Added the workspace folder name to the resulting path, with a small hash of the full path to avoid collisions. In my case it generates the following 2 paths:
```
/mnt/ch/ClickHouse/build/./cargo/build/workspace_24e5f
/mnt/ch/ClickHouse/build/./cargo/build/ffi_a7435
```

PR #614, opened 2025-04-16
Fix environment variables for custom target triples

### LINKED ISSUE #613: [Bug]: Corrosion sets wrong environment variables for custom target triples
### Current Behavior

The environment variables like `CARGO_TARGET_<triple>_LINKER` contain the full path (e.g. `/path/to/my_triple.json`) instead of just the filestem `my_triple`.

### Expected Behavior

Cargo expects just the filestem

### Steps To Reproduce

_No response_

### Environment

```markdown
- OS:
- CMake:
- CMake Generator:
```

### CMake configure log with Debug log-level

_No response_

### CMake Build step log

_No response_

jschwe:
Fixes the target triple name we use when setting some environment variables for Rust and cc-rs, when custom target triples are used.

Fixes #613


jschwe:
Closing in favor of #633

PR #633, opened 2025-08-25
fix: compiler env vars for cc-rs when using custom target triple

audunhalland:
Fixes (at least one symptom of) #613.

Might have the same effect as #614, unsure why that's still open. At least this is my version of the fix that works for for me, feel free to close if the other approach is better.


audunhalland:
It fixed something because `_CORROSION_RUST_TARGET_UNDERSCORE` refers to a full path to a json file in my case (with dashes converted into underscores), so it's useless as an env var generator here.

What's actually needed, and what I think was the original intention, is to pass to cc env vars of the form `CC_the_target_triple=/path/to/compiler`.

Issue #584, opened 2024-12-14
Building using Xcode generator


TheVitya:
Hello,

First of all, I want to say that I really like the project; it’s incredibly helpful to me.

I am trying to build a simple library using the Xcode generator. I haven’t seen any remarks or notices suggesting that corrosion does not support this setup.

However, when I try to build it using the build.sh script, I encounter a "Permission Denied" error. The same error occurs when I attempt to build it directly in Xcode.

![Screenshot 2024-12-14 at 13 44 37](https://github.com/user-attachments/assets/3cb644a2-88db-48e7-a54a-118139b94556)

I suspect this issue might be related to the combination of Xcode and Cargo because running the scripts manually from the terminal does not result in the same error.

I’m not sure what else to try at this point, as I’ve explored nearly every idea that came to mind.

I would greatly appreciate it if you could point me in the right direction to resolve this issue.

Thank you in advance for your help!

Repo: https://github.com/TheVitya/corrosion

jschwe:
Do you have the same problem when using Xcode and building for mac?
Corrosion does test in CI that using the XCode Generator works, but I'm not sure if anyone has used corrosion to build for iOS before.

TheVitya:
The initial error seems to be fixed, the generated SDKROOT was a bit faulty, but I think that was my fault. Now when I am trying to build it I get the following error: `ld: library 'System' not found`. I have seen that there is a related issue to that, but the paths that are mentioned there seems good. I think you would get the same error if you try it, the [repo](https://github.com/TheVitya/corrosion) is updated.

cmake: 3.31.2
cargo: 1.83.0
mac: 14.5 (23F79)
rustc: 1.83.0

jschwe:
I currently don't have any time to investigate this myself, but PRs are welcome. As I mentioned, I'm not aware of anyone using corrosion for iOS development right now, so this might require some investigation.

> ld: library 'System' not found

You might need to do something similar [as here for MacOS < 13](https://github.com/corrosion-rs/corrosion/blob/b88ec09915a6b12d1aaec7e451ecf0e41a831030/cmake/Corrosion.cmake#L665) and set `LIBRARY_PATH`.

TheVitya:
[...]

PR #635, opened 2025-08-28
Reset path in the environment for iOS builds

raphaelmenges:
Resolves linking issues when using Corrosion for iOS as target and macOS as host. Replicates a fix [originally suggested](https://github.com/TheVitya/corrosion/commit/ad77f27e8cc0e1bc59caca58c46e73843064cb41) by @TheVitya. See #584 for a detailed discussion.

I actually checked the value of my path environment variable before both suggested resets via: 

```
COMMAND
    ${CMAKE_COMMAND} -E echo "Current PATH: $ENV{PATH}"
```
It output both times `/usr/bin:/bin:/usr/sbin:/sbin`.

I do not understand why resetting the path to only `/usr/bin` fixes the issue. I am happy for suggestions and further knowledge to perhaps provide a more "proper" fix.

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -822,6 +822,7 @@ function(_add_cargo_build out_cargo_build_out_dir)
         # Build crate
         COMMAND
             ${CMAKE_COMMAND} -E env
+                "$<$<STREQUAL:${CMAKE_SYSTEM_NAME},iOS>:PATH=/usr/bin>"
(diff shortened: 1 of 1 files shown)


raphaelmenges:
According to Claude AI, `/bin`, `/usr/sbin`, and `/sbin` are managed by the OS and should / are usually be not used by users. Perhaps we should just remove these from the path for both commands and leave the path otherwise as it was at invocation?

PR #636, opened 2025-08-28
hack: building for iOS 

jschwe:
Posting to share the patches

### DIFF
--- a/cmake/Corrosion.cmake
+++ b/cmake/Corrosion.cmake
@@ -765,6 +765,9 @@ function(_add_cargo_build out_cargo_build_out_dir)
         list(APPEND corrosion_cc_rs_flags "AR_${_CORROSION_RUST_CARGO_TARGET_UNDERSCORE}=${CMAKE_AR}")
     endif()
 
+    # TODO: How to get a working host compiler properly? configure a dummy cmake project for the host?
+    list(APPEND corrosion_cc_rs_flags "CARGO_TARGET_AARCH64_APPLE_DARWIN_LINKER=/usr/bin/cc")
+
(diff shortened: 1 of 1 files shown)


cmake/Corrosion.cmake
```diff
@@ -765,6 +765,9 @@ function(_add_cargo_build out_cargo_build_out_dir)
         list(APPEND corrosion_cc_rs_flags "AR_${_CORROSION_RUST_CARGO_TARGET_UNDERSCORE}=${CMAKE_AR}")
     endif()
 
+    # TODO: How to get a working host compiler properly? configure a dummy cmake project for the host?
```
raphaelmenges, line 768: I like the idea to use a dummy cmake project for detecting the host compiler! I guess this dummy project would be generated via Make instead of XCode to not fall into the same trap of using the wrong `cc`?

jschwe: If you don't configure CMake to compile for iOS, I'm pretty sure XCode wouldn't setup the PATH with a `cc` for iOS, but provide one to compile for macos.

chocolacula:
Hey everyone, do you need any help with this PR? To be hones I tested it accidentally in my cross platform project. The platforms we support:
- aarch64-apple-darwin
- aarch64-apple-ios
- aarch64-apple-ios-sim
- x86_64-apple-ios
- aarch64-linux-android
- armv7-linux-androideabi
- i686-linux-android
- x86_64-linux-android
- wasm32-unknown-emscripten

And ios builds were failing due to incorrect system library(master branch). With corrosion from this branch the build is successful but a warning appears about provided -lSystem second time. So it looks working only some polishing needed I guess.

Issue #637, opened 2025-09-29
[Bug]: Exported impls for cxxbridge result in circular static library dependencies


AndrewLipscomb:
### Current Behavior

https://github.com/AndrewLipscomb/corrosion/tree/exported-impl-issue/test/cxxbridge/cxxbridge_exported_impls shows a test case situation from a larger project. 

We have a number of modules we try to expose under different bridges as a means of organising the codebase. When using cxx types over modules - you have to do https://cxx.rs/extern-c++.html#explicit-shim-trait-impls to produce the implementations of the template types. 

With the code in there - you end up with a circular dependency between the produced two static libraries. Under the CMake the final link command becomes 

```
/usr/bin/c++ -Wl,--dependency-file=CMakeFiles/test_main.dir/link.d CMakeFiles/test_main.dir/main.cpp.o -o test_main  libcxxbridge.a librust_lib.a -lgcc_s -lutil -lrt -lpthread -lm -ldl -lc
```

This fails as `libcxxbridge.a librust_lib.a` will fail to find some required symbols - inverting the order also means some required symbols will not be found. 

This can be somewhat sidestepped by adding in guards for recursive searching 

```
/usr/bin/c++ -Wl,--dependency-file=CMakeFiles/test_main.dir/link.d CMakeFiles/test_main.dir/main.cpp.
o -o test_main -Wl,--start-group librust_lib.a libcxxbridge.a -Wl,--end-group -lgcc_s -lutil -lrt -lpthread -lm -ldl -lc
```

But that puts a lot of onus on the final callee to use the right `target_link_library` flags. 

For this case alone - producing an `OBJECT` lib for `cxxbridge` instead of a `STATIC` lib solves the issue. However this appears to break the other `cpp2rust` test cases in a way I don't yet understand (this is the deepest I've ever gone into link issues) 

I was originally going to make a PR converting that library type to an `OBJECT` type but clearly there are more usecases going on here (or implementation specifics I don't quite understand yet). I'd be keen to get some other opinions on the right solution here. 

This is all under Linux on an x86 compiler if that is relevant in anyway. Reproduced on a few versions of GCC both on Ubuntu and Arch Linux. 

### Environment

```markdown
- OS:
- CMake:
- CMake Generator:
```

AndrewLipscomb:
[...]

PR #639, opened 2025-09-29
cxxbridge libraries now can hide the circular linking nature of the cxxbridge and rust code

AndrewLipscomb:
Based off investigation from https://github.com/corrosion-rs/corrosion/issues/637

The key issue here, from CXX's [docs](https://cxx.rs/build/other.html#linking-the-c-and-rust-together), is

> The generated C++ code and the Rust code generated by the procedural macro **both depend on each other.** Simple examples may only require one or the other, but in general your linking will need to handle both directions. For some linkers, such as LLD, this is not a problem at all. **For others, such as GNU ld, flags like --start-lib/--end-lib may help**.

Adds a new exported library from `corrosion_add_cxxbridge` called `{cxx_name}-link` to be an interface library hiding the potential circular linking nature of the cxxbridge and rust code from downstream linkers. If you do not meet the requirements in CMake to be able to generate this library - we issue a warning.


AndrewLipscomb:
Did a little bit of testing with the clang I have on my machine

Setting `CC=/usr/bin/clang` and `CXX=/usr/bin/clang++` and removing the circular link guards produced a similar error to GCC

```
/usr/bin/clang++  -I/home/andrew.lipscomb/code/corrosion/test/cxxbridge/cxxbridge_exported_impls/build/corrosion_generated/cxxbridge/cxxbridge/include -std=gnu++11 -MD -MT CMakeFiles/test_main.dir/main.cpp.o -MF CMakeFiles/test_main.dir/main.cpp.o.d -o CMakeFiles/test_main.dir/main.cpp.o -c /home/andrew.lipscomb/code/corrosion/test/cxxbridge/cxxbridge_exported_impls/main.cpp
[100%] Linking CXX executable test_main
/usr/bin/cmake -E cmake_link_script CMakeFiles/test_main.dir/link.txt --verbose=1
/usr/bin/clang++ -Xlinker --dependency-file=CMakeFiles/test_main.dir/link.d CMakeFiles/test_main.dir/main.cpp.o -o test_main  libcxxbridge-static.a librust_lib.a -lgcc_s -lutil -lrt -lpthread -lm -ldl -lc
/usr/bin/ld: librust_lib.a(rust_lib-a701772a4442be0b.4ueyecf3iactx3bvjofv2aa99.rcgu.o): in function `<rust_lib::bridge_b::ffi::NewVal as cxx::shared_ptr::SharedPtrTarget>::__new':
/home/andrew.lipscomb/code/corrosion/test/cxxbridge/cxxbridge_exported_impls/rust/src/bridge_b.rs:1:(.text._ZN84_$LT$rust_lib..bridge_b..ffi..NewVal$u20$as$u20$cxx..shared_ptr..SharedPtrTarget$GT$5__new17h2a66bdc89011d421E+0x25): undefined reference to `cxxbridge1$shared_ptr$NewVal$uninit'
/usr/bin/ld: librust_lib.a(rust_lib-a701772a4442be0b.4ueyecf3iactx3bvjofv2aa99.rcgu.o): in function `<rust_lib::bridge_b::ffi::NewVal as cxx::shared_ptr::SharedPtrTarget>::__get':
/home/andrew.lipscomb/code/corrosion/test/cxxbridge/cxxbridge_exported_impls/rust/src/bridge_b.rs:10:(.text._ZN84_$LT$rust_lib..bridge_b..ffi..NewVal$u20$as$u20$cxx..shared_ptr..SharedPtrTarget$GT$5__get17ha3cd68b6015dfbc6E+0xb): undefined reference to `cxxbridge1$shared_ptr$NewVal$get'
/usr/bin/ld: librust_lib.a(rust_lib-a701772a4442be0b.4ueyecf3iactx3bvjofv2aa99.rcgu.o): in function `<rust_lib::bridge_b::ffi::NewVal as cxx::shared_ptr::SharedPtrTarget>::__drop':
/home/andrew.lipscomb/code/corrosion/test/cxxbridge/cxxbridge_exported_impls/rust/src/bridge_b.rs:10:(.text._ZN84_$LT$rust_lib..bridge_b..ffi..NewVal$u20$as$u20$cxx..shared_ptr..SharedPtrTarget$GT$6__drop17hb9d5d9c0865c8905E+0x7): undefined reference to `cxxbridge1$shared_ptr$NewVal$drop'
```

Adding the link group back in resolved the issue. 

So - its not a GCC specific issue, as out-of-the-box Clang will not work either. 

Explicitly using LLD via `-DCMAKE_LINKER_TYPE=LLD` ended up linking successfully under clang 

```
[100%] Linking CXX executable test_main
/usr/bin/cmake -E cmake_link_script CMakeFiles/test_main.dir/link.txt --verbose=1
/usr/bin/clang++ -fuse-ld=lld -Xlinker --dependency-file=CMakeFiles/test_main.dir/link.d CMakeFiles/test_main.dir/main.cpp.o -o test_main  libcxxbridge-static.a librust_lib.a -lgcc_s -lutil -lrt -lpthread -lm -ldl -lc
```

Clang specifics on my Arch
```
andrew.lipscomb@archlinux:~$ clang --version
clang version 20.1.8
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

AndrewLipscomb:
I've redone this now to preserve the existing current behaviour - the new "linkable lib" is now under a `-link` suffix. I've updated the readme for that, as well as the tests. 

Again - probably worth going into the main branch too .... whats going on with the divergence between `main` and the stable/0.5 branch? I can rebase to `main` if that's better but it would be good to know what the release cycle looks like.

jschwe:
Thanks, I didn't have time to look at the PR yet (will try on the weekend), but some comments.

>> For others, such as GNU ld, flags like --start-lib/--end-lib may help.

Is there something prevent us from [specifying the cyclic dependency](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#cyclic-dependencies-of-static-libraries) `librust_lib.a libcxxbridge.a librust_lib.a`? It's been a while since I looked at this part, but I'm wondering if we could solve this without the need for rescanning, by just adding the libraries multiple times to the link line.

> So - its not a GCC specific issue, as out-of-the-box Clang will not work either.

gcc / clang / cc are just the driver for invoking the linker. The actual linker (e.g. ld vs lld) is more important (e.g. lld supports rescanning by default AFAIK).

> whats going on with the divergence between main and the stable/0.5 branch

Well, I've been planning to release the next release for a long time (1+ year) now (and I personally believe main is in a better condition than the stable branch). It's mainly that I was thinking of making the next version 1.0, but then there are a number of small things I would like to fix beforehand, and are just difficult to do due to time constraints. 
I would definitely recommend targeting (and perhaps also using) main though. And I should probably just make a release.

AndrewLipscomb:
> Is there something prevent us from [specifying the cyclic dependency](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#cyclic-dependencies-of-static-libraries) 

Didn't realise that was an option - I might have a play with that. It would be good to not foist a newer CMake on people if they are stuck in the past. 

>  lld supports rescanning by default AFAIK

Very correct, it does. My little experiment above showed that up. I don't really know what "real" projects do in terms of non-stock linkers, as all my systems use `ld` out of the box. I suppose supporting the "out of the box" situation best is the nicest thing to do for the next poor sucker looking at this :) 

> I would definitely recommend targeting (and perhaps also using) main though

I'll rebase when I get a chance to come back and fix the above points up. I might updating my internal tools to use main as well. That said - a release would be great too

AndrewLipscomb:
I have made a new PR with a rebased branch - https://github.com/corrosion-rs/corrosion/pull/640. Its solving the cyclic dependency via 2x `link_libraries` statements, and seems to translate to other Cmake targets. 

@jschwe up to you on what you'd want to do with this branch and PR. I'll close it for now - let me know if you'd want it reopened.

Issue #624, opened 2025-05-21
[Bug]: cmake/ninja targets from corrosion are always dirty


milianw:
### Current Behavior

Using corrosion in a cmake project will never go into the `nothing to do` state, there are always dirty targets.

To reproduce, first build your project that is using corrosion:
```
$ ninja
```

Now the expectation is that you'd get:

```
$ ninja
ninja: no work to do.
```

Instead you get (the example output below is from a tiny project, `image_filter` is the crate we try to import via corrosion):

```
$ ninja -t explain -v
ninja explain: output ImageFilter/cargo-prebuild_image_filter of phony edge with no inputs doesn't exist
ninja explain: output ImageFilter/CMakeFiles/_cargo-build_image_filter doesn't exist
ninja explain: ImageFilter/CMakeFiles/_cargo-build_image_filter is dirty
ninja explain: ImageFilter/cargo-prebuild_image_filter is dirty
ninja explain: ImageFilter/CMakeFiles/_cargo-build_image_filter.util is dirty
ninja explain: ImageFilter/_cargo-build_image_filter is dirty
ninja explain: ImageFilter/libimage_filter.a is dirty
ninja explain: imageviewer-app is dirty
ninja explain: ImageFilter/cargo-build_image_filter is dirty
ninja explain: output _deps/corrosion-build/all of phony edge with no inputs doesn't exist
ninja explain: _deps/corrosion-build/all is dirty
ninja explain: ImageFilter/all is dirty
[...]

Issue #206, opened 2022-08-30
Use cargo depfile to prevent calling cargo if not necessary


jschwe:
- add_custom_command takes an optional `DEPFILE` command
  - Generator support depends on CMake version
- would allow skipping calling cargo if not necessary
- Blocked on #201

swilcox3:
@jschwe This feature would be great for me.  I'm looking at how to make it so we don't need to call cargo when we do null builds.  Right now it looks like the actual build step is a `add_custom_target`, which runs every time no matter what.  I'm tinkering with your source right now to try to see if I can make it a `add_custom_command` and maybe pass in the generated depfile.  You've done fantastic work on this though, and you're definitely more of a CMake wizard than I am, so if you're able to support this officially that'd be great.

jschwe:
>  if I can make it a add_custom_command and maybe pass in the generated depfile

I think it should be possible. Some thoughts:

- `DEPFILE` is not supported by all CMake Generators, although I think the main ones all support it by now.
- The `INTERFACE_CORROSION_CARGO_PROFILE` target property support might need to be removed to get this feature working. `INTERFACE_CORROSION_CARGO_PROFILE` currently allows users to set the cargo build profile (release / debug / ...) per target, but is incompatible with specifying the cargo output build artifacts as an `OUTPUT` of the custom build command. Corrosion currently works around this by having the cargo build custom target not having any `BYPRODUCTS` and instead specifying the final byproducts in a `POST_BUILD` command of the custom target.
 I think the POST_BUILD command is also necessary and needs to be seperate from the main build, due to corrosion  supporting the `OUTPUT_DIRECTORY` CMake variables.

dtugend:
Sorry regarding my earlier comment claiming that we would need INTERFACE_CORROSION_CARGO_PROFILE, I found another way that your seem to support officially, that also seems to work, that I wasn't aware of, here is an example:

```cmake
include(FetchContent)
include(ExternalProject)

#
# Setup "corrosion" for including Rust code in our projects:
#

FetchContent_Declare(
    Corrosion
    GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git
    GIT_TAG 1499b14e4906a2890f5cee1547c8848db261753d # v0.6.1
)
FetchContent_MakeAvailable(Corrosion)

#
# Import our main Rust workspace:
#

if(${CMAKE_BUILD_TYPE_DEBUG})
[...]

PR #654, opened 2025-12-16
RFC: use cargo's depfiles

russelltg:
This is not complete yet, but wanted to gauge interest before polishing up. 

An annoying thing about corrosion is that it doesn't use ninja's dependency tracking at all, instead depending on cargo's. This means that cargo has to be invoked for every call of ninja, regardless of if anything needs to be done or not. 

Cargo outputs depfiles that are exactly for this purpose, so this PR hooks that up. This lets ninja do nothing when there is no work to do, but otherwise do work. 

This required temporarily disabling the hostbuild & profile features, as those are 'target' generator features which are not allowed in this context. 

Do you think it would be acceptable to replace the generator expressions with a DEFER call like is done for the byproducts?


russelltg:
See also https://github.com/corrosion-rs/corrosion/issues/624 and https://github.com/corrosion-rs/corrosion/issues/206

PR #656, opened 2025-12-30
Prepare v0.6.1 release

jschwe:
Release recent ios fixes


ogoffart:
Thanks for making the release.

BTW, the 0.6.1 release is not marked as latest, so the 0.6.0 still show as the release on the side of https://github.com/corrosion-rs/corrosion

image

jschwe:
Thanks - Should be fixed now!

PR #666, opened 2026-03-17
Experimental Crubit Integration

thunderseethe:
Background Context: I work on https://crubit.rs a C++/Rust interop tool and I'm trying to add CMake support for it. Crubit works by ingesting a `.rlib` file and generating a `.h` and `.rs` file that provide C++ bindings to the rust code within the `.rlib`. I would like to get Crubit working with CMake and I think Corrosion would make that much easier for the end user (and me as a CMake novice) given it already does a lot of the heavy lifting to integrate a Cargo project into a C++ project.

This PR introduces a new `corrosion_experimental_crubit($target_name)` function that toggles on Crubit for a library target in a crate. The intended workflow is to toggle on Crubit for your target and then call import_crate:

```
corrosion_experimental_crubit(rust_lib)
corrosion_import_crate(MANIFEST_PATH rust_lib/Cargo.toml)
```

Within Corrosion this flag modifies the behavior of `_corrosion_add_library_target`, `_add_cargo_build`, and `_generator_add_package_targets`. When Crubit is enabled for a target, `_generator_add_package_targets` will fetch Crubit, check if the cargo subcommand is installed, and install from the fetched sources if not. 

`_corrosion_add_library_target` checks if the target is a Crubit library and adds the generated header as a byproduct when it is. Because a Crubit target has the generated header, we specify the target_include_directories. It also adds a dependency on our support libraries, which are provided by the fetch in `_generator_add_package_targets`.

`_add_cargo_build` now toggles the command it runs based on if the target is a crubit library. When the target is not a Crubit lib, the normal flow executes and calls `cargo rustc`. When it is a Crubit library, we call `cargo cpp_api_from_rust` (the subcommand we installed earlier). Under the hood, this subcommand performs a `cargo build` and so takes a lot of the same flags as `cargo rustc` but in a sligtly different CLI.

I am a novice at CMake, so please advice on style/idiom stuff. I've tried to match the style corrosion uses, but there's definitely some of the generator expression stuff I am not understanding super well.


cmake/CorrosionGenerator.cmake
```diff
@@ -133,7 +133,60 @@ function(_generator_add_package_targets)
             continue()
         endif()
 
-        if("staticlib" IN_LIST kinds OR "cdylib" IN_LIST kinds)
+        get_property(crubit_enabled GLOBAL PROPERTY CORROSION_CRUBIT_${target_name})
+        if(crubit_enabled)
+            include(FetchContent)
+            FetchContent_Declare(
```
thunderseethe: I wanted to call this out as I wasn't really sure this was the right way to do this. I wanted to allow consumers to specify their own version of Crubit and fallback to grabbing the git repo if they do not. From the docs that sounds like what `FetchContent_declare` does? But it seems like it also needs to interact with `find_packages` (perhaps that isn't a consideration here since Crubit isn't an OS package like a C lib would be).

cmake/Corrosion.cmake
```diff
@@ -871,11 +909,71 @@ function(_add_cargo_build out_cargo_build_out_dir)
...
+            get_filename_component(rustc_bin_dir "${_CORROSION_RUSTC}" DIRECTORY)
+            get_filename_component(rustc_root_dir "${rustc_bin_dir}" DIRECTORY)
+            set(rustc_lib_dir "${rustc_root_dir}/lib")
+        endif()
+        set(rustc_lib_dir_env "${lib_path_env_var}=${rustc_lib_dir}")
+        cmake_path(GET ACB_CARGO_CRUBIT_BIN PARENT_PATH cargo_cmd_dir)
+        set(path_env_var "${cargo_cmd_dir}:$ENV{PATH}")
+
+        # Since cargo cpp_api_from_rust uses cargo build internally, we pass cargo build args after `--`
+        set(cargo_args
```
jschwe, line 927: Does it use `cargo build` or `cargo rustc`? Corrosion uses `cargo rustc`, to pass some rustflags to the top-level crate only, and it seems you are just dropping `local_rustflags_genex` arguments here? I guess not everything needs to work out of the box, but I think it should be documented what is explicitly not supported.

cmake/CorrosionGenerator.cmake
```diff
@@ -133,7 +133,60 @@ function(_generator_add_package_targets)
...
+                        )
+                    add_custom_target("_corrosion_cargo_crubit"
+                        DEPENDS "${cargo_crubit}"
+                        )
+                endif()
+            endif()
+        endif()
+
+        if(crubit_enabled AND ("lib" IN_LIST kinds OR "rlib" IN_LIST kinds))
+          set(is_crubit_lib ON)
```
jschwe, line 213: One question I have here: How well-defined is it to generate bindings on `lib` and `rlib` packages? When building e.g. a cargo shared-library object A that depends on an rlib B, isn't the linker free to eliminate any functions of B that are not reachable via A's public API?

cmake/Corrosion.cmake
```diff
@@ -884,23 +982,9 @@ function(_add_cargo_build out_cargo_build_out_dir)
...
-                --manifest-path "${path_to_toml}"
-                --target-dir "${cargo_target_dir}"
-                ${cargo_profile}
-                ${flags_genex}
-                # Any arguments to cargo must be placed before this line
-                ${local_rustflags_delimiter}
-                ${local_rustflags_genex}
+                "PATH=${path_env_var}"
+            ${cargo_bin}
+                ${cargo_args}
```
jschwe, line 985: What are the outputs of this command when using crubit? Does it still compile the crate and all dependencies? Does it differe between rlibs, and executables/shared libraries?

cmake/CorrosionGenerator.cmake
jschwe: FetchContent seems fine to me. But since you later want to do a `cargo install` on a crate, that is not on crates.io, and pretty unknown / niche I'm wondering if we might want to move this whole block into a function, that is only conditionally available (behind a cmake cache variable), so that it's very easy to verify that this crate can't be installed when just using corrosion normally.

test/crubit/rust_lib/src/lib.rs
```diff
@@ -0,0 +1,14 @@
+pub fn add(left: u64, right: u64) -> u64 {
+    left + right
+}
+
+#[cfg(test)]
```
jschwe: This block seems unnecessary.

test/crubit/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+cmake_minimum_required(VERSION 3.22)
+project(test_project VERSION 0.1.0)
+include(../test_header.cmake)
+
+corrosion_experimental_crubit(rust_lib)
+corrosion_import_crate(MANIFEST_PATH rust_lib/Cargo.toml)
+
+add_executable(main main.cpp)
+target_link_libraries(main PRIVATE rust_lib)
```
jschwe: So what magic does crubit do internally that this works? Does it create a staticlib or a shared library (essentially overriding the actual library type)?

jschwe (commented):
Sorry for taking so long to respond. I took a first look, and still have a couple of question about how crubit works.

thunderseethe:
Thanks for the response. I wanted to give you the heads up. I'm OOO until the 17th but I will get back to you after that.

cmake/Corrosion.cmake
thunderseethe: `cargo cpp_api_from_rust` only runs on a lib/rlib target today. It generates C++ bindings to the public API of the corresponding target, so it isn't super meaningful to run it on a binary target or a target that is already producing a C compatible staticlib/sharedlib.

The outputs will be:

  * `include/` a directory containing a generated header with the bindings for each crate in the graph that received bindings (generally this will be the current crate and all dependencies).
  * `lib_rust_lib.a` a staticlib containing the definitions for the declarations in the headers. These will be the extern C decls we generate to cross the FFI boundary and the compiled rust code that implements the underlying functions.
  
  `cargo cpp_api_from_rust`, at a high level, does three things:
  
  * builds your crate using `cargo build`
  * Runs our bindings generator over all the `rlib`s produced by `cargo build`
    * This produces the generated headers and generates rust code containing the extern C decls
  * Compiles the generated rust (which depends on the crate) into a staticlib. 
  
  That was an info dump so please let me know if I can clarify anything more.

test/crubit/CMakeLists.txt
thunderseethe: I touched on this above, but our subcommand walks the definitions contained in the `rlib` and generated corresponding C++ definitions and a suite of extern C functions to cross the FFI boundary. The definitions get put in generated headers (~ one per crate) and the generated rust containing the extern C functions get compiled into a staticlib that is linked into the `rust_lib` target.

cmake/Corrosion.cmake
thunderseethe: Today the subcommand uses `cargo build`. I don't think anything prevents using `rustc`. One minor issue our cargo subcommand takes it's own args, so where `cargo rustc` can use `<cargo args> -- <rustc args>` our subcommand has `<subcommand args> -- <cargo args>`. But I can find a way to workaround so that the local rustflags can be specified.

I will take a look at updating the subcommand and circle back here.

test/crubit/rust_lib/src/lib.rs
thunderseethe: Done.

cmake/CorrosionGenerator.cmake
thunderseethe: I've moved this over to a new function. I originally had the function definition conditioned on a cache variable, but that caused ordering issues (the definition would see the cache variable as off but the function call would see it as on), so I've instead opted to unguard the definition but the function call is guarded on the cache variable.

cmake/CorrosionGenerator.cmake
thunderseethe: I believe this is true but will only matter at the linking stage. I don't think it will be the case that the rlib itself is paired down knowing that A will only need a subset of it. I believe one of:

1) A would only lazily generate the object code for what they need out of B
2) A would generate object code for B and then the linker would do DCE for that object code.

What we're doing is definitely off the usual usecase for rlibs. They're intended to be more so an implementation detail of rustc and an intermediary. They're moving towards being a stable platform you can build tooling on with things like https://github.com/rust-lang/project-stable-mir

cmake/Corrosion.cmake
thunderseethe: Actually I forgot what this section of code was doing. For a crubit target we combine local and global flags into the global flags genex and specify them all via the environment variable on the cargo command. Will that work as intended or do the local args need to be specified on the CLI (vs the env var)?

thunderseethe:
@jschwe I think I have responded to all the comments and update the code accordingly. Please take another look.

cmake/Corrosion.cmake
jschwe: > One minor issue our cargo subcommand takes it's own args, so where cargo rustc can use <cargo args> -- <rustc args> our subcommand has <subcommand args> -- <cargo args>

Couldn't another `--` delimiter solve this? E.g. `<subcommand args> -- <cargo args> -- <rustc args>`?

> Will that work as intended or do the local args need to be specified on the CLI (vs the env var)?

Global flags affect all crates, while local flags only affect the top-level crate. In many cases it probably will not make a difference, but in some it could. In practice it shouldn't be a major issue I hope.
I think the main reason corrosion opted for the passing flags via `cargo rustc`, was that corrosion mainly needs to set linker options (for the top-level crate), and adding flags to the "correct / active" RUSTFLAGS method is not trivial (there are like 5+ different ways to set RUSTFLAGS)

cmake/Corrosion.cmake
jschwe: To clarify: Is `lib_rust_lib.a` self-contained like a regular cargo built staticlib, or does it only contains the API boundary and depends on the `rlib`s for actual implementation (thus allowing to link multiple of these static-libs into one C++ binary / library?

thunderseethe: The first one. We generate a rust file that `pub use rust_lib::*` and contains a whole host of generated ffi functions, but then give that to rustc and compile it to a staticlib like it was any other rust code. And on the C++ side the only thing we add to the build graph are the generated header and that staticlib.

cmake/Corrosion.cmake
thunderseethe: Really great point and I was not understanding the distinction between global and local. I updated the subcommand to use `rustc` and update the args here to pass `-- ${local_rustflags}` which will get threaded through to that rustc call verbatim.

thunderseethe:
Whoops sorry. I don't know how git works.

jschwe:
Would you or someone else working on / familiar with crubit by chance happen to be at rustweek?

thunderseethe:
Yes! Taylor Cramer, giving one of the [keynotes](https://2026.rustweek.org/talks/taylor/), is attending and leads the Crubit team. She would be happy to talk about it.

cmake/Corrosion.cmake
```diff
@@ -1141,6 +1235,22 @@ function(corrosion_import_crate)
     endif()
 endfunction()
 
+#[=======================================================================[.md:
+ANCHOR: corrosion-experimental-crubit
+** EXPERIMENTAL **: This function is currently still considered experimental
+  and is not officially released yet. Feedback and Suggestions are welcome.
+
```
jschwe, line 1231: It would probably be very useful to have a link to the official crubit documentation here, and maybe an example that convinces a user skimming the corrosion docs to click the link and learn more about crubit.

cmake/Corrosion.cmake
```diff
@@ -338,12 +339,21 @@ function(_corrosion_copy_byproduct_deferred target_name output_dir_prop_names ca
...
     message(DEBUG "Adding command to copy byproducts `${file_names}` to ${dst_file_names}")
+    get_property(crubit_enabled GLOBAL PROPERTY CORROSION_CRUBIT_${target_name})
+    # Crubit produces an `include/` directory of generated headers that needs to be copied (rather than a single file).
+    if (crubit_enabled AND (output_prop_name STREQUAL "LIBRARY_OUTPUT_DIRECTORY"))
+      set(cli_command "copy_directory_if_different")
+      # We need to add our directory name to the output for copy_directory_if_different.
+      set(output_dir "${output_dir}/${file_names}")
+    else()
+      set(cli_command "copy_if_different")
+    endif()
```
jschwe: I believe for cbindgen / cxxbridge we tell the tools directly where we want the files to be generated to. Would it make sense to directly tell crubit into which directory we want the include files to be generated to, instead of copying?

cmake/Corrosion.cmake
```diff
@@ -871,11 +915,74 @@ function(_add_cargo_build out_cargo_build_out_dir)
...
+            --package ${package_name}
+            ${cargo_profile}
+            ${flags_genex}
+            # Any arguments to cargo must be placed before this line
+            ${local_rustflags_delimiter}
+            ${local_rustflags_genex}
+        )
+    else()
+        set(rustc_lib_dir_env "")
+        set(path_env_var "$PATH")
```
jschwe: This would also need to be `$ENV{PATH}` to read from the environment. 
But I'd prefer if you could make setting the variable optional, so for the non-crubit build it isn't set, i.e do something like `"${path_var}"` whit path_var being either `PATH=....` or empty.

cmake/Corrosion.cmake
```diff
@@ -871,11 +915,74 @@ function(_add_cargo_build out_cargo_build_out_dir)
...
+              set(lib_path_env_var "LD_LIBRARY_PATH")
+            endif()
+            # On Linux, shared objects are typically in the lib directory
+            get_filename_component(rustc_bin_dir "${_CORROSION_RUSTC}" DIRECTORY)
+            get_filename_component(rustc_root_dir "${rustc_bin_dir}" DIRECTORY)
+            set(rustc_lib_dir "${rustc_root_dir}/lib")
+        endif()
+        set(rustc_lib_dir_env "${lib_path_env_var}=${rustc_lib_dir}")
+        cmake_path(GET ACB_CARGO_CRUBIT_BIN PARENT_PATH cargo_cmd_dir)
+        set(path_env_var "${cargo_cmd_dir}:$ENV{PATH}")
```
jschwe: Note: on windows the path seperator is not `:`, but `;`. You could use `cmake_path(CONVERT ... TO_NATIVE_PATH_LIST)` or check again for CMAKE_HOST_WIN32 like you did above.

cmake/CorrosionGenerator.cmake
```diff
@@ -47,6 +47,57 @@ function(_cargo_metadata out manifest)
...
 
+function(_corrosion_setup_crubit out_cargo_crubit out_crubit_depends)
+    if(NOT CORROSION_EXPERIMENTAL_CRUBIT)
+        return()
+    endif()
+    include(FetchContent)
+    FetchContent_Declare(
+        Crubit
+        GIT_REPOSITORY https://github.com/google/crubit.git
+        GIT_TAG main
```
jschwe: Will you be publishing tags? Since crubit is still under development, I think it would be better to avoid depending on a moving target like `main`, since that could lead to broken builds and non-reproducibility. How about specifying a specific commit, but allowing the user to override the tag / version? That would give reproducible behavior by default, but also allow users experimenting to quickly update.

cmake/CorrosionGenerator.cmake
```diff
@@ -47,6 +47,57 @@ function(_cargo_metadata out manifest)
...
+
+        if(NOT TARGET "_corrosion_cargo_crubit")
+            file(MAKE_DIRECTORY "${local_crubit_install_dir}")
+
+            message(STATUS "Building cargo-cpp_api_from_rust from Crubit")
+            add_custom_command(OUTPUT "${cargo_crubit}"
+                COMMAND ${CMAKE_COMMAND}
+                -E env
+                "CARGO_BUILD_RUSTC=${_CORROSION_RUSTC}"
+                ${_CORROSION_CARGO} install
```
jschwe: I think you should prefer using 
```
"CARGO_BUILD_RUSTC=$CACHE{CORROSION_TOOLS_RUSTC}"
                    $CACHE{CORROSION_TOOLS_CARGO}
```

which aligns with what cxxbridge does. This is mainly important when cross-compiling, since we always want to compile crubit for the host platform.

cmake/CorrosionGenerator.cmake
```diff
@@ -47,6 +47,57 @@ function(_cargo_metadata out manifest)
...
+    if(NOT CORROSION_EXPERIMENTAL_CRUBIT)
+        return()
+    endif()
+    include(FetchContent)
+    FetchContent_Declare(
+        Crubit
+        GIT_REPOSITORY https://github.com/google/crubit.git
+        GIT_TAG main
+    )
+    FetchContent_MakeAvailable(Crubit)
```
jschwe, line 60: If I see it correctly, then `crubit_support` is made available by Crubit. Perhaps we could check if the target is defined after this, to catch it if Crubit renames the target.

I'm also wondering if we could configure Crubit with a flag to only run the CMake code which is necessary for us (e.g. skip the LLVM related configuration if thats possible). Otherwise we would want to document the prerequisites and impact on configure time in the documentation.

cmake/Corrosion.cmake
```diff
@@ -483,6 +506,9 @@ function(_corrosion_add_library_target)
         add_library(${target_name}-static STATIC IMPORTED GLOBAL)
         add_dependencies(${target_name}-static cargo-build_${target_name})
         set_target_properties(${target_name}-static PROPERTIES COR_FILE_NAME ${static_lib_name})
+        if (is_crubit_lib)
+          set_target_properties(${target_name}-static PROPERTIES COR_HEADER_FILE_NAME ${header_name})
```
jschwe: Is this a leftover? it doesn't seem to be read anywhere.

jschwe (commented):
General questions: After this is merged, would there be some commitment from your side to be available for maintaining the crubit code and further improving the integration?

cmake/CorrosionGenerator.cmake
```diff
@@ -133,7 +184,26 @@ function(_generator_add_package_targets)
...
+                )
+            endif()
+        endif()
+        if(crubit_enabled AND ("lib" IN_LIST kinds OR "rlib" IN_LIST kinds))
+          set(is_crubit_lib ON)
+        else()
+          set(is_crubit_lib OFF)
+        endif()
+
+        if("staticlib" IN_LIST kinds OR "cdylib" IN_LIST kinds OR is_crubit_lib)
```
jschwe, line 218: I don't think it is needed to import dependency crate as a cmake target. At least the build rule itself should not be needed (I'm thinking of lib A is a top-level rlib, lib B is a dependency of lib A, and the user wants to generate bindings to lib B, why would we need to import lib B, since the artifact we would be using would be `libA.a`. headers, should probably also be associated with the target of the static-lib. Could you clarify, why you believe importing dependencies as a cmake target is needed? If it's just about being able to import a top-level `rlib` target, then we should be able to do that with the existing CRATE_TYPES override.

cmake/Corrosion.cmake
thunderseethe: Yes it was thank you.

cmake/Corrosion.cmake
thunderseethe: Done.

cmake/Corrosion.cmake
thunderseethe: Good point. Done.

cmake/Corrosion.cmake
thunderseethe: 😅  I sure should've thought of that. Great point. Done.

cmake/CorrosionGenerator.cmake
thunderseethe: Added a check for `crubit_support`.

Could you tell me a little bit more about the LLVM related configuration? Crubit depends on LLVM for the Rust calling C++ direction of interop, but for the work here on C++ calling Rust we don't depend (directly) on LLVM and I don't think we're building LLVM in CMake. The binary we build depends on the llvm component of a Rust toolchain installation and the generated C++ code relies on being built with Clang.

cmake/CorrosionGenerator.cmake
thunderseethe: Good catch. Done.

cmake/CorrosionGenerator.cmake
thunderseethe: I agree I don't think we need cmake targets for dependencies. My imagined workflow is you call corrosion (with crubit enabled) pointing at a Cargo project containing one lib target. The generated CMake target would include the directory of generated headers and the single generated static lib containing the compiled crate graph of the cargo project. On the C++ side you should have access to everything you need for any dependency between the headers and staticlib.

It's very possible I've done something silly here due to not understanding what's available in the CMake/Corrosion already. My original understanding was that `CRATE_TYPES` allows for override to set one of `staticlib`, `cdylib`, or `bin`. Would I specify `CRATE_TYPES` for the corresponding target when someone calls `corrosion_experimental_crubit`? Or ask users to specify it in `corrosion_import_crate` when they enable crubit?

cmake/Corrosion.cmake
thunderseethe: It is definitely unfortunate how I mangled this function to add the copy directory branch. For the staticlib we generate, I don't think we can do much to adjust where it's output. It's produced by a `cargo rustc` call which I believe always puts it in the target directory.

We have some more wiggle room with the headers as crubit's subcommand controls where those live. However, I think it makes sense for it to put it's output's in the target directory as well, given it's invoked as a cargo subcommand. AFAICT, cbindgen/cxxbridge are both evoked as independent CLIs and just happen to be called on a Cargo project. WDYT?

cmake/CorrosionGenerator.cmake
thunderseethe: I would like to get us to a point where we have a meaningful versioning scheme, but today we are quite far from that. I think before we could consider something like that we would have to get the other direction working in open source.

That being said you are right leaving this hanging isn't wise. I'll see if I can get a tag setup such that we can update it on our end without having to change this code everytime we roll forward.

thunderseethe:
> General questions: After this is merged, would there be some commitment from your side to be available for maintaining the crubit code and further improving the integration?

Yes I plan to continue maintaining this code moving forward. I expect that after landing this initial setup we'll see a flurry of bugs as folks begin to exercise it and I'm committed to fixing those and anything else that comes up.

cmake/CorrosionGenerator.cmake
jschwe: If you don't want to  publish tags / versions yet, we could also pin commits here.

cmake/Corrosion.cmake
jschwe: Is the staticlib location any different then it would be for a regular build? I'm mainly trying to avoid adding / complicating code where perhaps no (new) code would also work fine. Particular this copying code has been quite a pain so I'd prefer avoiding changes if they aren't necessary.

> However, I think it makes sense for it to put it's output's in the target directory as well, given it's invoked as a cargo subcommand.

Is there any use-case for those headers being in the target directory? We are integrating cargo into CMake here, so I think letting CMake tell cargo where to put something is reasonable. If we gained the ability to tell cargo where to but build artifacts, I would also use that instead of copying from the target dir.

cmake/CorrosionGenerator.cmake
jschwe: >My imagined workflow is you call corrosion (with crubit enabled) pointing at a Cargo project containing one lib target. The generated CMake target would include the directory of generated headers and the single generated static lib containing the compiled crate graph of the cargo project. On the C++ side you should have access to everything you need for any dependency between the headers and staticlib.

That sounds reasonable.

> My original understanding was that CRATE_TYPES allows for override to set one of staticlib, cdylib, or bin. Would I specify CRATE_TYPES for the corresponding target when someone calls corrosion_experimental_crubit? Or ask users to specify it in corrosion_import_crate when they enable crubit?

Does crubit respect the `rustc` flags telling rustc which crate type to build, or does it always build a static lib? 
As a user I would like to decide if I'm building a static or a shared library, and CRATE_TYPES is the current mechanism for users to override the artifact type from `rlib` to something else.

cmake/CorrosionGenerator.cmake
thunderseethe: I've added a latest tag in the repo that's pinned to a commit with green CI. That should provide some stability vs always living at head while allowing us to update without having to update the `.cmake` files here. I also imagine most users will pin their own version of Crubit manually.

cmake/Corrosion.cmake
thunderseethe: I see what you're saying. I've swapped the headers to be placed via an `--out-dir` CLI flag. We still place the staticlib in the target directory and copy if different, but this reuses the existing copy byproduct code as is, so hopefully that's alright. The code to copy the header directory is now cleaned up.

cmake/CorrosionGenerator.cmake
thunderseethe: Today it always builds a staticlib. I don't think there's anything technically preventing us from producing any of the crate types. In theory, we could produce an rlib for our generated crate containing the extern C decls. Granted, I don't know what one would do that.

That being said, I am skeptical of the use case for a non-staticlib as the final output. With how our binding generation is structured, mainly around how we name symbols, it's unlikely that a user could keep one header and swap out multiple dylibs under the hood. They would need to regenerate both the header and underlying library to keep names in sync. We use rustc's name mangling, today, so this is true across toolchain updates as well, which is another situation I imagine one might want dylibs.

I think we are free down the line to add support for dylibs if we find there is a use case for them I'm not considering. Until then I'd like to punt on the feature for simplicity. WDYT?

thunderseethe:
Okay I believe I've updated the code to respond to the latest round of comments and followed up on all the threads. PTAL when you have some time @jschwe (sorry if I missed one; I'm definitely struggling to navigate github PR UI).

thunderseethe:
Okay I updated the tests to use a similar structure to cxx/bindgen where they're disabled by default and conditionally enabled in a specific job. Hopefully tests should be passing now, but I think I cannot retrigger CI?

jschwe:
@thunderseethe anything we can do about the dependency on a specific rustc version?
At least for CI testing it would be really good to have this be deterministic (both in terms of which crubit version is pulled, and what rustc versions that's compatible with).

thunderseethe:
Yeah this was not working as intended. I introduced the `latest` tag on our end to track the last green CI but that got broken when we started running zizmor over our github actions. I'm taking a look to get that tag fixed up and then we should be able to avoid breakages like this in the future by defaulting to `latest` which will use a rustc that works

PR #668, opened 2026-04-21
Honor CARGO_HOME env var when searching for rust binaries

milianw:
Don't just check $HOME/.cargo/bin, first check $CARGO_HOME/bin when looking for binaries. This way we can also find cargo in custom install locations. It should also allow us to simplify packaging of corrosion in vcpkg, by adding CARGO_HOME to VCPKG_KEEP_ENV_VARS.

### DIFF
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -224,7 +224,7 @@ set(Rust_RESOLVE_RUSTUP_TOOLCHAINS ON CACHE BOOL ${_RESOLVE_RUSTUP_TOOLCHAINS_DE
 # This block checks to see if we're prioritizing a rustup-managed toolchain.
 if (DEFINED Rust_TOOLCHAIN)
     # If the user specifies `Rust_TOOLCHAIN`, then look for `rustup` first, rather than `rustc`.
-    find_program(Rust_RUSTUP rustup PATHS "$ENV{HOME}/.cargo/bin")
+    find_program(Rust_RUSTUP rustup PATHS "$ENV{CARGO_HOME}/bin" "$ENV{HOME}/.cargo/bin")
     if(NOT Rust_RUSTUP)
         if(NOT "${Rust_FIND_QUIETLY}")
             message(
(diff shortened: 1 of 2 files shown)


cmake/FindRust.cmake
```diff
@@ -224,7 +224,7 @@ set(Rust_RESOLVE_RUSTUP_TOOLCHAINS ON CACHE BOOL ${_RESOLVE_RUSTUP_TOOLCHAINS_DE
 # This block checks to see if we're prioritizing a rustup-managed toolchain.
 if (DEFINED Rust_TOOLCHAIN)
     # If the user specifies `Rust_TOOLCHAIN`, then look for `rustup` first, rather than `rustc`.
-    find_program(Rust_RUSTUP rustup PATHS "$ENV{HOME}/.cargo/bin")
+    find_program(Rust_RUSTUP rustup PATHS "$ENV{CARGO_HOME}/bin" "$ENV{HOME}/.cargo/bin")
```
jschwe: Seems reasonable, but probably we should check if `CARGO_HOME` is defined, otherwise the first PATH would collapse to `/bin`. Maybe add an `if DEFINED ... use CARGO_HOME/bin else $HOME/.cargo/bin`?

cmake/FindRust.cmake
```diff
@@ -253,10 +253,10 @@ else()
             return()
         endif()
     else()
-        find_program(_Rust_COMPILER_TEST rustc PATHS "$ENV{HOME}/.cargo/bin")
+        find_program(_Rust_COMPILER_TEST rustc PATHS "$ENV{CARGO_HOME}/bin" "$ENV{HOME}/.cargo/bin")
         if(NOT EXISTS "${_Rust_COMPILER_TEST}")
             cmake_path(CONVERT "$ENV{HOME}/.cargo/bin" TO_CMAKE_PATH_LIST _cargo_bin_dir)
-            set(_ERROR_MESSAGE "`rustc` not found in PATH or `${_cargo_bin_dir}`.\n"
+            set(_ERROR_MESSAGE "`rustc` not found in PATH, `$CARGO_HOME/bin` or `${_cargo_bin_dir}`.\n"
```
jschwe, line 272: Wouldn't the `$` need to escaped?

milianw: I thought so too, but I tried it and it worked. AFAIU you'd need to write `${...}` to get expansion here

cmake/FindRust.cmake
milianw: do you care strongly about this? Note that `HOME` might also be unset (and actually is, in a vcpkg-environment). The paths will just fail to lookup and that's fine then, no?

but I can change it if you feel strongly, personally I'd argue that it's not needed and actually makes the code a bit more messy than needed. We basically would need:

```
set(_rustup_paths_hints)
if (DEFINED ENV{CARGO_HOME})
    list(APPEND _rustup_paths_hints "$ENV{CARGO_HOME}/bin")
endif()
...
find_program(Rust_RUSTUP rustup PATHS ${_rustup_paths_hints})
```

jschwe: > The paths will just fail to lookup and that's fine then, no?

`/bin` can exist (on unix systems), and then we would be looking there first instead of in the user home first, so the priority would change.

jschwe (approved):
Thanks for the PR, I pushed some minor changes and will merge once CI passes.

